PR #49, opened 2017-08-09
Fix heading links in 'Jobs to be done' documentation page

### LINKED ISSUE #47: 3rd level header links in dev docs have colliding titles
Visit https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#prediction-service and you will notice that the 3rd level headings are the same in each section, this prevents you from navigating to any section other than the 1st.

## Expected Behavior
Clicking on a heading (e.g., Interface API) should go to the appropriate section

## Current Behavior
You can only navigate within the first section

## Possible Solution

Use different heading names, or the [`:ref:` directive](https://stackoverflow.com/a/19543591/408556)

## Acceptance criteria

- [ ] Clicking on a heading (e.g., Interface API) results in a navigation to the appropriate section

foo-bar-baz-qux:
## Description
Changed headings on the "Jobs to be done" documentation page so that the heading links do not collide. Heading links are generated from the text of the headings, so having the same heading text on the same page causes issues. This seems to be the simplest solution.

I initially tried looking for other solutions, like `:ref:`, but the markdown to RST parser generates links automatically from the text with no obvious way to override.

Since the TOC is auto-generated using these heading links, it also becomes an issue when trying to link from the navigation pane to any `:ref` links on the page. Could manually create the TOC and navigation but would be tedious to update in the future.

Also tried enabling `autostructify` and trying to use the references in native RST embedded using `eval_rst` code blocks along with the section titles, but the section titles seem to cause confusion for the parser.

Fixes #47.

## Motivation and Context

Allows independent heading links on the "Jobs to be done" documentation page.

## How Has This Been Tested?

Built the sphinx documentation locally and made sure the links updated and worked.

## CLA

### DIFF
--- a/docs/design-doc.md
+++ b/docs/design-doc.md
@@ -232,11 +232,11 @@ Task areas: `IMPORTING`
 
 ![](_static/wireframe-0-open.svg)
 
-#### Prediction service
+#### Imagery - Prediction service
 
 RESERVED
 
-#### Interface API
+#### Imagery - Interface API
 
 1.  List available DICOM images on local filesystem.
 2.  Display such details as may be available (e.g. metadata) for any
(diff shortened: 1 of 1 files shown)


lamby:
Looks good to me   :+1: 

Thanks for trying ``autostructify``  :)  :100:

lamby:
Merged; many thanks :)

Issue #38, opened 2017-08-04
Create a smaller test image dataset


reubano:
## Overview

Currently test images are stored in the [git repository](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/tests/assets). These images are formatted as directories of [DICOM](https://en.wikipedia.org/wiki/DICOM) files. Each directory contains hundreds of files, and consumes ~100MB of space. We would like to reduce the number of files to the bare minimum needed to detect nodules and pass the tests.

## Expected Behavior

The number of DICOM directories and images is reduced to the minimum amount necessary (1 - 3 directories, each with 5 - 10 DICOM files).

## Technical details

 - Each directory contains an xml file with metadata about the included DICOM files. You may need to edit this file. 

## Acceptance criteria

 - [ ] number of test images is reduced as described above
 - [ ] the tests suite passes with the reduced image dataset

*NOTE: All PRs must follow the standard PR checklist.*

tdraebing:
I had a look into this and found that the dimensions metadata is in the metadata of each *.dcm-file. The xml-file denotes the position of the marked nodules and is not needed to load the images. 
`pydicom` actually does not care if you delete images of a series, but `dicom-numpy` does, since it tests, whether the distance between slices is the same. Since the numbering of the files is not in the same order as the z-position of the slice, simply deleting files does not work here.
I wrote a function that can reduce the dimensions of a series in x, y and z. I think that this function may be useful in general, because new test image sets might be needed for different tasks. I will create smaller test sets asap and create a pull request. Should I include the script as well in this pull request?
What should happen to the full data sets? I would propose to change the docker script to only include the small dataset, when running the server, but keep the original files in the repo.

reubano:
Yes, you can include the script. Also, feel free to remove the original files from the repo. The goal of the PR is to reduce the repo's total size.

tdraebing:
I finally finished the pull request.
But I have some concerns that I wanted to share before the pull request is merged. While the reduced size of the dataset is no issue for testing tooling or similar tasks, it might reduce the dataset's information content considerably, which might cause issues for training models etc. A lot of positional information gets lost for example: Where in the lung is the nodule/tumor? Also the nodules annotated in the *.xml-files are pretty wide spread, so we lost a few here.
[...]

PR #57, opened 2017-08-17
Reducing test set size and adding a script to crop dicom images.

tdraebing:
Replaced large test images with smaller ones to decrease repo size.

## Description

I decreased the size of all three data sets by reducing the number of images (cropping in z-dimension). For the data sets LIDC-IDRI-0001 and LIDC-IDRI-0003 I chose the regions that contain a big mass inside the lung, which are probably tumors. LIDC-IDRI-0002 does not contain such a mass. Thus I chose a range of slices that contained nodules as stated by the *.xml file. (Disclaimer: I am not a medical professional. As a biologist I only have very limited knowledge about diagnosing lung cancer on CT-images). The *.xml files were removed from the data set for now. The data sets are in the same location as the original data sets. The names changed to be more declarative and now represent the depth of the slices in mm instead of a non-consecutive id.
Additionally a script was added to `prediction/src/preprocess/crop_dicom.py` that allows to crop dicom series in x-, y- and z-dimensions for future use.
Existing tests were updated accordingly.

Issue #38 

## Motivation and Context

The datasets were quite large. Smaller datasets would substentially decrease the repo size. The created script might be useful for preprocessing in prediction tasks.

## How Has This Been Tested?

Unit tests for the new script were created to test whether the expected output is created. Existing tests were updated to appropriately deal with the smaller test sets (One test relied on a *.xml-file being in the folder). The other tests resulted in no errors. Smaller datasets can be opened using an external DICOM-Viewer (RadiAnt).

## CLA


reubano:
Thanks @tdraebing! Nice job with this. One question initially: would you mind using [Google style formatting](https://github.com/concept-to-clinic/concept-to-clinic/blob/c484acf1af419235ee98889332c1c890a35a140c/prediction/src/preprocess/load_dicom.py#L39-L43) for the docblocks?

tdraebing:
I updated the pull request. The doc string now follows Google style formatting. Also the full test set was included again next to the cropped dataset. I updated the docker-configuration to load the small dataset into the `/images`-folder when running tests.

prediction/src/preprocess/crop_dicom.py
```diff
@@ -0,0 +1,50 @@
...
+def crop_dicom(path_to_dicom, begin, end, output=None):
+    """
+    This function crops a dicom series in the x-, y- and z-dimension. If an output path is provided, it will save the
+    new series at that location.
+
+    Examples for how to specify cropping mask using the begin- and end-parameter:
+
+        The given DICOM series contains images with the resolution of 512x512 pixels and 133 images
+        starting at a depth of -10 mm to a depth of -340 mm. Using begin = [0, 0, -10] and end = [512, 512, -340] would
+        give you back the full series. Using begin = [100, 200, [0, 0, -160] and end = [120, 220, -170] would give you a
```
reubano: `[100, 200, [0, 0, -160]` is this a typo?

prediction/src/tests/test_crop_dicom.py
```diff
@@ -0,0 +1,35 @@
...
+    cropped_series = cd.crop_dicom(dicom_path, [100, 100, -101], [120, 120, -106])
+    assert len(cropped_series) == 2
+    assert cropped_series[0].Rows == 20
+    assert cropped_series[0].Columns == 20
+    assert np.asarray(cropped_series[0].pixel_array).shape[0] == 20
+    assert np.asarray(cropped_series[0].pixel_array).shape[1] == 20
+    assert np.asarray(list(filter(lambda x: x.SliceLocation == -102.5, cropped_series))[0].pixel_array)[0][0] == \
+        np.asarray(list(filter(lambda x: x.SliceLocation == -102.5, uncropped_series))[0].pixel_array)[100][100]
+    assert np.asarray(list(filter(lambda x: x.SliceLocation == -102.5, cropped_series))[0].pixel_array)[19][19] == \
+        np.asarray(list(filter(lambda x: x.SliceLocation == -102.5, uncropped_series))[0].pixel_array)[119][119]
```
reubano: I think we can replace `list(filter(lamba x...`'s here with list comprehensions, e.g., `[x for x in cropped_series if x.SliceLocation == -102.5]`

reubano:
Made a couple comments... afterwards, a rebase + squash should be all that's needed. Thanks!

tdraebing:
Thanks for reviewing! Should be fixed now.

reubano:
Quick question, does the small test data set use the actual images or the `git lfs` ones? It should be the actual images, but I just ran a test and noticed the files on my computer are still linked via `lfs`.

tdraebing:
I am not very familiar with `lfs`, but as I understand it, right now `lfs` is scanning for all `*.dcm`-files in the repo-folder as specified in the `.gitattributes`-file. Thus the small dataset  would also be handled by `lfs`. `git lfs ls-files` indeed shows that the small datasets are managed by `lfs`.
I am not completely sure how to fix it correctly without doing a mistake. I guess I would have to follow this guide: https://help.github.com/articles/removing-files-from-git-large-file-storage/ and change the entry in `.gitattributes` to a more specific path. But as I understand it, this will completely remove the files from the repo. Is changing the `.gitattributes`-file and committing with the image files maybe enough to replace the `lfs`-pointers in the repo with the proper files? Could anybody comment on that?

reubano:
Ok... maybe a fix can bundled in with #59? @lamby @pjbull

PR #55, opened 2017-08-16
Document how to update documentation

musale:
## Description
I have bootstrapped the steps needed for local installation, building and editing of the documentation for this project.

## Reference to official issue
This addresses the issue #41 

## Motivation and Context
This change enables developers to be able to edit the documentation locally. Since docker is the official development configuration, I used it to create the application and link it to the current development.

## How Has This Been Tested?
Locally, when you run

`$ docker-compose -f local.yml up`

the documentation application will be running on (http://localhost:8002/)[http://localhost:8002/]

## CLA
- [x] I have signed the CLA; if other committers are in the commit history, they have signed the CLA as well

PR #61, opened 2017-08-21
#8 Add Sphinx Autodoc 

WGierke:
## Description
This pull request should add the ability to auto-generate code documentation based on Google-styled docstrings. 

This should solve #8 

## Motivation and Context
Instead of manually adapting the documentation to the written code and its docstrings, we can now auto-generate the Sphinx documentation based on the docstrings in the code.

## How Has This Been Tested?
1. I merged my code into #55 (so it's easier for me to update the documentation :-) )
2. I started docker-compose
3. I saw the auto-summaries of all modules, classes and methods that have proper docstrings
4. To confirm that it really supports Google-style docstrings, I downloaded [this](http://www.sphinx-doc.org/en/stable/ext/example_google.html) sample code having lots of Google-docstrings
5. I restarted docker-compose
6. I confirmed that the docstrings of that file were added to the Sphinx documentation

Overview
screenshot from 2017-08-22 01-12-20
Details of the interface.backend package
screenshot from 2017-08-22 01-12-35
One legacy method that already has a complete docstring
screenshot from 2017-08-22 01-13-10
The Google-docstring file I added for testing purposes was rendered correctly
screenshot from 2017-08-22 01-04-16

## CLA

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -7,6 +7,10 @@ SPHINXBUILD   = sphinx-build
 SPHINXPROJ    = ConcepttoClinic
 SOURCEDIR     = .
 BUILDDIR      = _build
+APIDOCDIR_INTERFACE    = _apidoc_interface
+APIDOCDIR_PREDICTION   = _apidoc_prediction
+SPHINXAPIDOC  = sphinx-apidoc
+PROJDIR       = ../
 
 # Put it first so that "make" without argument is like "make help".
 help:
@@ -18,3 +22,12 @@ help:
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+.PHONY: html
+html:
+	$(SPHINXAPIDOC) -f -o $(APIDOCDIR_INTERFACE)/ $(PROJDIR)/interface/
+	$(SPHINXAPIDOC) -f -o $(APIDOCDIR_PREDICTION)/ $(PROJDIR)/prediction/
+	@echo
+	@echo "apidoc build finished. The api source files are in $(APIDOCDIR)/"
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+	@echo "Sphinx build finished. The build files are in $(BUILDDIR)/"
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,6 @@ prediction/data/*
 
 # ignore data
 data/
+
+# ignore Sphinx apidoc
+_apidoc_*
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,16 +45,7 @@ extensions = [
 ]
 
 napoleon_google_docstring = True
-napoleon_numpy_docstring = True
-napoleon_include_init_with_doc = False
-napoleon_include_private_with_doc = False
-napoleon_include_special_with_doc = True
-napoleon_use_admonition_for_examples = False
-napoleon_use_admonition_for_notes = False
-napoleon_use_admonition_for_references = False
-napoleon_use_ivar = False
-napoleon_use_param = True
-napoleon_use_rtype = True
+napoleon_numpy_docstring = False
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,15 +20,14 @@
 import os
 import sys
 
-DOCS_DIR = os.path.abspath(__file__)
-PROJECT_DIR = os.path.join(DOCS_DIR, os.path.pardir)
+DOCS_DIR = os.getcwd()
+PROJECT_DIR = os.path.abspath(os.path.join(DOCS_DIR, os.pardir))
 
 INTERFACE_DIR = os.path.join(PROJECT_DIR, 'interface')
 sys.path.insert(0, INTERFACE_DIR)
 
 PREDICTION_DIR = os.path.join(PROJECT_DIR, 'prediction')
 sys.path.insert(0, PREDICTION_DIR)
-
 
 # -- General configuration ------------------------------------------------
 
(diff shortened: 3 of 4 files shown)


isms:
@WGierke FWIW, while we diagnose #63 your build is failing on this line from `test_docker.sh`:
```
$ sh tests/test_docker.sh
+docker-compose -f local.yml run prediction pytest
Creating network "concepttoclinic_default" with the default driver
Creating volume "concepttoclinic_postgres_data_dev" with default driver
============================= test session starts ==============================
platform linux -- Python 3.6.2, pytest-3.1.3, py-1.4.34, pluggy-0.4.0
rootdir: /app, inifile:
collected 4 items / 1 errors 
==================================== ERRORS ====================================
_________________ ERROR collecting src/tests/test_endpoints.py _________________
ImportError while importing test module '/app/src/tests/test_endpoints.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
src/tests/test_endpoints.py:13: in <module>
    from src.factory import create_app
E   ModuleNotFoundError: No module named 'src'
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
=========================== 1 error in 0.72 seconds ============================
The command "sh tests/test_docker.sh" exited with 2.
```

WGierke:
Thanks a lot @isms that clearly helps :)

isms:
@WGierke This may be a little less painful if you can run the commands in `test_docker.sh` locally - here's the broken part.
```
$ flake8 prediction
prediction/src/tests/test_endpoints.py:10:1: E402 module level import not at top of file
prediction/src/tests/test_endpoints.py:11:1: E402 module level import not at top of file
prediction/src/tests/test_endpoints.py:12:1: E402 module level import not at top of file
prediction/src/tests/test_endpoints.py:14:1: E402 module level import not at top of file
prediction/src/tests/test_endpoints.py:15:1: E402 module level import not at top of file
```

WGierke:
Thanks a lot @isms , I was able to fix everything :)

prediction/src/tests/test_endpoints.py
```diff
@@ -4,13 +4,13 @@
...
 from functools import partial
 import json
-
 import pytest
-
-from flask import url_for
-from src.factory import create_app
+import sys
+sys.path.append('.')
+from src.factory import create_app  # NOQA
```
pjbull: Hm, it's strange for this test to begin failing with this change.

It's probably because the change adds `__init__.py`s in `prediction` and `interface`. Two things to try instead of changing `sys.path`:
 (1) Does doing a relative import work? (e.g. `from .factory`)
 (2) Does importing `prediction.src.factory` work?

Other than that, we can also look at the call to `pytest`. IIRC there are a couple ways to do that which change how imports work for the test cases.

WGierke: Hey @pjbull thanks for your feedback!
I tried out your suggestions but that leads to
1) `ModuleNotFoundError: No module named 'app.src.tests.factory'`
2) `ModuleNotFoundError: No module named 'prediction'`

I'm not that familiar with `pytest` but I can do some research about that.

pjbull: Thanks, also just to confirm, do the tests pass locally if you remove the `__init__.py` files?

WGierke: Removing `interface/__init__.py` and `prediction/__init__.py` makes the tests still pass but Sphinx unable to detect interface and prediction as package so it won't be able to automatically summarize the classes using their docstrings.

isms: What about just tweaking the `Makefile` to add a step `touch [whatever]/__init__.py` then making the docs then removing them?

isms:
@WGierke If you're making a series of commits, would you mind pushing only once you've finished a chunk of work? The flurry of small pushed commits are jamming up the Travis build queue!

Also don't forget to rebase onto master one you're ready for a review :+1:

WGierke:
@isms oh I'm sorry. I wasn't able to see that as I have no access to the Travis site. Are you able to cancel them?
Would another possibility be to squash the commits when merging the branch?

lamby:
> I wasn't able to see that as I have no access to the Travis site.

This has been filed as https://github.com/concept-to-clinic/concept-to-clinic/issues/68

prediction/src/tests/test_endpoints.py
WGierke: Very good idea, thanks @isms ! [That](https://github.com/concept-to-clinic/concept-to-clinic/pull/61/files#diff-68a7f9edd4ffbca6adb99e2c165cfd52R31) should implement it this way

docs/Makefile
```diff
@@ -17,4 +20,21 @@ help:
...
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+.PHONY: clean
+clean:
+	rm -rf $(BUILDDIR)/*
+
+.PHONY: html
+html:
+	# Make interface and prediction to packages so apidoc can detect them
+	touch interface/__init__.py
```
lamby: Problem here is that if we fail to complete the build, we leave these files lying around in the repo. This is both messy, but also a problem in that even if we ``.gitignore``d them, as they are now packages this is going surely going to cause some grief wrt tests or.. something. Anyway, gut feel is that this is wrong. :/

docs/interface/backend.api.rst
```diff
@@ -0,0 +1,61 @@
+backend\.api package
```
lamby: How come we need to "escape" the dots?

WGierke:
@lamby Sorry for the delay, I finally figured out how to solve it. Now sphinx-apidoc should be used correctly and the paths in [conf.py](https://github.com/concept-to-clinic/concept-to-clinic/pull/61/files#diff-85987f48f1258d9ee486e3191495582dR23) should also be fixed.
screenshot from 2017-08-25 19-17-50
screenshot from 2017-08-25 19-18-17

docs/Makefile
```diff
@@ -7,6 +7,10 @@ SPHINXBUILD   = sphinx-build
 SPHINXPROJ    = ConcepttoClinic
 SOURCEDIR     = .
 BUILDDIR      = _build
+APIDOCDIR_INTERFACE    = _apidoc_interface
+APIDOCDIR_PREDICTION   = _apidoc_prediction
```
WGierke: Multiple apidoc directories are necessary since for each package a 'modules.rst' is created, which would be overwritten by the last execution of `sphinx-apidoc` when the output should only be saved in one directory.

lamby:
Thanks @WGierke !

Issue #14, opened 2017-08-04
Pass calculated summary statistics from segment model to API


pjbull:
## Overview
Once we produce summary statistics for tumors, those need to be returned to the frontend through the API.

## Expected Behavior
Currently, the `segment` endpoint only returns a path to a file which has a binary mask of the DICOM image. This is useful for displaying the nodules, but we also want to report summary statistics. This issue tracks taking both the summary statistic calculations and the path to the binary masks generated by `segment.trained_model.predict` and surfacing both of those pieces of information through the API.

Design doc reference:
[Jobs to be done > Detect and select > Prediction service](https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#prediction-service)

## Technical details
 - This feature should be implemented in [views.py](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/prediction/src/views.py) for the prediction application and wherever else down the stack is necessary to pass along the data 

## Related issues

Depends on #13, which tracks actually calculating the volumes. It's acceptable to create a stub for that work and implement this independently. You may need to read #13 to fully understand this issue.

## Acceptance criteria
 - [ ] An API endpoint for nodule summary statistics
 - [ ] documentation for the summary statistics endpoint

*NOTE: All PRs must follow the standard PR checklist.*

datu925:
Hi @pjbull! Acceptance criteria for this issue seems to be a carryover from the identification model.

pjbull:
Hey @datu925!! Good to see you here! I've updated those, good catch. 🙇

reiinakano:
Interested in this issue but am a bit unclear on exact details of the expected API.

Should I modify the return result of `POST segment/predict` to include summary statistics? Or should I create a new endpoint, say `GET/POST segment/statistics`, that take the binary mask path and outputs the volumes?

PR #65, opened 2017-08-23
Adjust segment model interface to return volumes of each nodule

reiinakano:
## Description
To solve issue #14, I went with simply modifying the output of `/segment/predict/` to a dictionary of the format:
```python
{'binary_mask_path': str,
 'volumes': list[float]}
```

For Issue #14 

## How Has This Been Tested?
I changed the unit test for `test_segment` to test for the new format of the returned response.

## CLA

### DIFF
--- a/prediction/src/algorithms/segment/trained_model.py
+++ b/prediction/src/algorithms/segment/trained_model.py
@@ -27,7 +27,32 @@ def predict(dicom_path, centroids):
              'z': int}
 
     Returns:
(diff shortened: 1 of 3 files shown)


reiinakano:
As I'm unclear on the exact API you are looking for, I went with what, for me, made most sense. Will wait for your feedback before I start the documentation for the endpoint.

lamby:
Great stuff again. Looks good to me.. @reubano ? :)

reubano:
🚢

Issue #10, opened 2017-08-04
Self-document the prediction API endpoint


pjbull:
## Overview
A useful API will tell you how to use it. Update the API endpoints to be more helpful!

## Expected Behavior
Currently, there is just a placeholder if you `GET` our prediction API endpoints. We expect this call to tell us what the endpoint does, what the required parameters are, and what it returns.

To do this, we'll need to update our response payload to have that information.

## Technical details
 - This feature should be implemented in the [`prediction/src/views.py`](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/prediction/src/views.py#L70-L74)
 - For each algorithm, we should add the right: `description`, `expected_parameters`, `return_values`
 - This feature should have tests for each algorithm in [`prediction/src/tests/test_endpoints.py`](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/prediction/src/views.py#L70-L74)

## Acceptance criteria
 - [ ] accurately documents all three prediction endpoints
 - [ ] has passing tests

*NOTE: All PRs must follow the standard PR checklist.*

lamby:
Closed in https://github.com/concept-to-clinic/concept-to-clinic/pull/66. Thanks !

PR #66, opened 2017-08-23
Self-documenting endpoint

reiinakano:
## Description

This is a very simple fix, I simply changed the endpoint to return the *docstring* of the selected `predict` function. Personally, I think this is the best way to keep it "self-documenting" and keeping the description in one place only. Let me know if you have problems with this approach.

Also, I added an `__init__.py` file to the `algorithms` folder to make it a valid Python package.

#10 

## How Has This Been Tested?

Modified the `test_endpoint_documentation` test to make sure that calls return the appropriate docstring.

## CLA

### DIFF
--- a/prediction/src/tests/test_endpoints.py
+++ b/prediction/src/tests/test_endpoints.py
@@ -11,6 +11,9 @@ import pytest
 
 from flask import url_for
 from src.factory import create_app
+from src.algorithms import classify
+from src.algorithms import identify
+from src.algorithms import segment
 
 
 def get_data(response):
@@ -48,10 +51,22 @@ def test_endpoint_documentation(client):
     r = client.get(url)
     data = get_data(r)
 
-    assert 'description' in data
+    assert data['description'] == identify.trained_model.predict.__doc__
+
+    url = client.url_for('predict', algorithm='segment')
+    r = client.get(url)
+    data = get_data(r)
+
+    assert data['description'] == segment.trained_model.predict.__doc__
+
+    url = client.url_for('predict', algorithm='classify')
+    r = client.get(url)
+    data = get_data(r)
+
+    assert data['description'] == classify.trained_model.predict.__doc__
 
 
-def test_indentify(client):
+def test_identify(client):
     url = client.url_for('predict', algorithm='identify')
 
     test_data = dict(dicom_path='')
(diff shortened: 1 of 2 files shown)


reiinakano:
Travis test failed due to Git LFS issues probably.
```
warning: Clone succeeded, but checkout failed.

You can inspect what was checked out with 'git status'

and retry the checkout with 'git checkout -f HEAD'
```

prediction/src/tests/test_endpoints.py
```diff
@@ -11,6 +11,9 @@
 
 from flask import url_for
 from src.factory import create_app
+from src.algorithms import classify
```
lamby: You can write this more cleanly as:

```
from src.algorithms import classify, identify, segment
```

prediction/src/tests/test_endpoints.py
```diff
@@ -48,10 +51,22 @@ def test_endpoint_documentation(client):
...
+
+    url = client.url_for('predict', algorithm='classify')
+    r = client.get(url)
+    data = get_data(r)
+
+    assert data['description'] == classify.trained_model.predict.__doc__
 
 
-def test_indentify(client):
+def test_identify(client):
```
lamby: :D

prediction/src/tests/test_endpoints.py
```diff
@@ -48,10 +51,22 @@ def test_endpoint_documentation(client):
...
-    assert 'description' in data
+    assert data['description'] == identify.trained_model.predict.__doc__
+
+    url = client.url_for('predict', algorithm='segment')
+    r = client.get(url)
+    data = get_data(r)
+
+    assert data['description'] == segment.trained_model.predict.__doc__
+
+    url = client.url_for('predict', algorithm='classify')
```
lamby: Feels like we could at least put this into a loop or a test helper instead of duplicating the code twice.

reiinakano:
@lamby Fixed. But test failed again due to LFS issues

lamby:
@reiinakano Thanks. Lets see how those tests do before merging, but please squash your commits so we don't have an ugly "fix tests" commit ;)

reiinakano:
Hi @lamby, Github has a one click option to squash all commits when merging to master. It's what most open source projects do now. Wouldn't that be enough? This branch will be deleted after merging anyway. All that'll be left is a one nice commit on the master branch history.

Issue #11, opened 2017-08-04
Update error handling in the prediction API


pjbull:
## Overview
We need to adapt the Data Science Bowl algorithms to produce possible centroid locations for nodules within an image rather than just `P(cancer)` for the whole image.

## Expected Behavior
Currently, we just catch all exceptions and ignore them. We want to return as part of the json payload some useful information about what went wrong. For example, if the request didn't have the right parameters, we say what was missing. If the DICOM image could not be found, we say the file was not where we expected it.

## Technical details
 - This feature should be implemented in the [`prediction/src/views.py`](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/prediction/src/views.py#L94-L95) file
 - We probably need multiple `except ExceptionType` blocks for different kinds of expected errors. For example:

```
    try:
       ...
    except ImportError as ie:
        # one error message
    except ValueError as e:
        # a different error message
```

 - We need a catch-all error message at the end
 
## Out of scope

We don't need to catch every possible exception. This issue is to identify how some common errors might occur and then return useful error messages.

## Acceptance criteria
 - [ ] returns useful error messages depending on the exception
 - [ ] tests for the failure cases that are handled

*NOTE: All PRs must follow the standard PR checklist.*

PR #67, opened 2017-08-23
Improve error handling in prediction API

alexanderrich:
Prediction API now returns message specifying if a needed parameter is not
passed to the prediction algorithm, and passes error messages that occur within
the prediction algorithm.

## Description
The API passes an error message if a non-existent algorithm is selected, as before. The API now also passes an error message if a necessary parameter to the algorithm is missing, specifying the algorithm chosen and the required parameter.  All other errors (e.g., bad DICOM file path) should be caught within the prediction algorithm, and these messages are caught and passed through the API.

Fixes issue #11 

## How Has This Been Tested?
These changes are tested through additional tests in test_endpoints.py. To properly test the response to bad DICOM paths, I modified the "predict" functions of the three "trained_model.py" modules to load the passed DICOM path. This loaded image is currently simply discarded after loading.

## CLA

### DIFF
--- a/prediction/src/views.py
+++ b/prediction/src/views.py
@@ -57,14 +57,15 @@ def predict(algorithm):
         algorithm (str): The prediction algorithm to use. One of 'segment',
             'classify', or 'identify'.
     """
-    # empty string to track if there was an
-    error = ''
 
     # dictionary to hold the response
     response = dict()
 
+    # string to contain error message
+    error = ""
+
     if algorithm not in PREDICTORS:
-        errormsg = 'Error! {} is not a valid algorithm. Please choose from {}.'
+        errormsg = "Error! '{}' is not a valid algorithm. Please choose from {}."
         error = errormsg.format(algorithm, set(PREDICTORS))
 
     # describe API on GET
@@ -84,17 +85,19 @@ def predict(algorithm):
             prediction = predict_method(**payload)
 
             response.update({
-                'prediction': prediction
+                'prediction': prediction,
             })
 
-        except Exception:
-            error = "There was an error with your request."
+        except Exception as e:
+            # pass errors from prediction function along with function chosen
+            error = "Error using algorithm '{}': {} ({})."
+            error = error.format(algorithm, str(e), type(e).__name__)
 
-    # set the status code and messages for the response
+    # set the status code for the response
     if error:
         response.update({
             'error': error,
-            'status': 500
+            'status': 500,
         })
     else:
         response.update({
--- a/prediction/src/algorithms/classify/trained_model.py
+++ b/prediction/src/algorithms/classify/trained_model.py
@@ -7,6 +7,8 @@
     for if nodules are concerning or not.
 """
 
+from src.preprocess.load_dicom import load_dicom
+
 
 def predict(dicom_path, centroids):
     """ Predicts if centroids are concerning or not.
@@ -33,6 +35,7 @@ def predict(dicom_path, centroids):
              'z': int,
              'p_concerning': float}
     """
+    load_dicom(dicom_path)
     for centroid in centroids:
         centroid['p_concerning'] = 0.5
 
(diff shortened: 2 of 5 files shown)


prediction/src/views.py
```diff
@@ -81,29 +89,27 @@ def predict(algorithm):
...
+        except KeyError as e:
+            errormsg = 'Algorithm \'{}\' requires missing parameter {}'.format(algorithm, str(e))
+            return compose_error(errormsg)
+        except Exception as e:
+            # pass other errors from prediction function
+            return compose_error(str(e))
+
+    # set the status code for the response
+    response.update({
+        'status': 200
```
lamby: You've dropped a trailing ',' here ;)  This helps with avoiding diff noise if someone adds a new line :)

alexanderrich: Ah, that makes sense. I'd noticed those trailing commas but wasn't sure if they were intentional. I'll add them back in!

reubano:
Current travis error

window 2017-08-25 at 14 56 18

reubano:
Hi @alexanderrich good job with this! A few requests before we can get this merged:

  1. rebase and fix the conflicts
  2. figure out what's causing the travis test to fail
  3. refactor the `predict` function in `views.py` to `return` just once

num 2 looks related to `git lfs` so it may be out of your control. I'll try to investigate that one...

alexanderrich:
Great, I'll fix (1) and (3). (2) seems connected to Issue #69 - I noticed that Travis failed after my first and third commits, but ran successfully on my second, so I think the issue is external.

alexanderrich:
(1) and (3) are resolved, but it looks like Travis is encountering the same issue.

prediction/src/views.py
```diff
@@ -81,20 +82,27 @@ def predict(algorithm):
...
 
             response.update({
-                'prediction': prediction
+                'prediction': prediction,
             })
 
-        except Exception:
-            error = "There was an error with your request."
+        except KeyError as e:
+            error = 'Algorithm \'{}\' requires missing parameter {}'.format(algorithm, str(e))
```
lamby: This is a little ugly :)  Why not:

```
error = "Algorithm '{}' requires missing parameter {}'.format(algorithm, str(e))
```
? :)

lamby:
LGTM. @reubano ? :)

alexanderrich:
Good point, just made that quotation-mark fix and squashed with previous commit

lamby:
@alexanderrich Thanks. I'll leave this assigned over to @reubano to merge...

prediction/src/views.py
```diff
@@ -81,20 +82,27 @@ def predict(algorithm):
         try:
             predict_method = PREDICTORS[algorithm]
 
-            prediction = predict_method(**payload)
```
reubano: I think the `**payload` version was a bit cleaner

reubano:
> ...it looks like Travis is encountering the same issue.

Yea, hopefully we'll get that addressed soon.

prediction/src/views.py
alexanderrich: Good point, I've switched to using the kwargs method. In the process it seemed like it would just be simpler to catch all errors in the same way, simply stating the name of the algorithm used and then printing the resulting error. Let me know what you think of this change.

reubano: My only worry is that by removing the `KeyError` check, the error message may not be as meaningful. I.e., `KeyError: 'dicom_path'` vs `missing parameter 'dicom_path'`.

reubano:
@alexanderrich I believe the travis errors have been fixed. A rebase to master should make everything turn green :).

prediction/src/views.py
alexanderrich: errors from wrong parameters should now be TypeErrors, not KeyErrors, which are a bit more verbose (e.g., the error message specifies whether your call to predict() has an unneeded parameter or an extra parameter). I also added the type of the error to the end of the message, which could be a useful extra piece of information at least during development.

reubano:
All green! You mind squashing to 1 commit?

lamby:
@reubano wrote:
> All green! You mind squashing to 1 commit?

Just a reminder we can do this on merge.. even within Github!

Issue #29, opened 2017-08-04
Show current Git version in top-level navigation


lamby:
## Overview

Interface should include the current Git version in the top-right corner. This is in order to ease reporting of issues; any screenshots will implicitly include the version, potentially saving wasted effort when debugging possibly-fixed bugs, etc.

## Expected Behavior

In the interface, the truncated SHA of the Git version used to build the site should be displayed, eg. ``b1f2ad46``.

## Technical details

It should also ensure it works in production where the ``.git/`` metadata directory will not exist, thus some change to the deployment scripts may be required to capture this data.. Care should be taken re. caching to not inflict a performance penalty on every page load.

## Acceptance criteria

The top-right corner will display the SHA-1.

## Blocked by
Issues: #9

PR #71, opened 2017-08-24
#29 Show current Git version in top-level navigation

WGierke:
For debugging purposes it would be really nice to see in the top right corner which version is currently used. The version is the short SHA-1 commit hash.

## Description
Since I didn't want to add the whole `.git` directory to the Docker container, I just added `.git/logs/HEAD` as a volume which contains historical information about branch changes etc. 

This should solve #29 

## Motivation and Context
The big advantage is that by just adding `HEAD` we don't have the full Git history in our container. Furthermore, every time we switch branches or commit, the `HEAD` file is synchronized with the container (as well as the Docker volumes). In the container, a Django view parses the short current commit hash out of that file and displays it. This way, when switching branches not only the code in the container is synchronized with the code at the host but also the commit hash is updated accordingly. The container can then also be deployed to production.

## How Has This Been Tested?
1. `docker-compose -f local.yml up`  yields
screenshot from 2017-08-24 23-11-18

(with the short version of the commit of this branch: a9f0e6a)
2. add an empty commit: `git commit -m "l" --allow-empty`
3. I refreshed the page and ended up with
screenshot from 2017-08-24 23-12-13
which was the short version of the commit hash of the empty commit (1bd5c33)

## CLA

### DIFF
--- a/interface/backend/static/views.py
+++ b/interface/backend/static/views.py
@@ -1,0 +1,7 @@
+from django.shortcuts import render
+
+
+def home(request):
+    current_hash = open('/HEAD', 'r').readlines()[-1].split(' ')[1]
+    short_hash = current_hash[:7]
+    return render(request, '../frontend/index.html', {'short_hash': short_hash})
--- a/interface/backend/static/views.py
+++ b/interface/backend/static/views.py
@@ -1,7 +1,6 @@
+from django.conf import settings
 from django.shortcuts import render
 
 
 def home(request):
-    current_hash = open('/HEAD', 'r').readlines()[-1].split(' ')[1]
-    short_hash = current_hash[:7]
-    return render(request, '../frontend/index.html', {'short_hash': short_hash})
+    return render(request, '../frontend/index.html', {'short_hash': settings.APP_VERSION_NUMBER})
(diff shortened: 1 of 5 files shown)


WGierke:
Any feedback is highly appreciated, especially since I'm a complete novice concerning Docker :)

interface/backend/static/views.py
```diff
@@ -0,0 +1,7 @@
+from django.shortcuts import render
+
+
+def home(request):
+    current_hash = open('/HEAD', 'r').readlines()[-1].split(' ')[1]
```
lamby: So, there is "prior art" within Django for this kind of lookup - I think it's called ``get_version`` or similar. Basically, it needs to a) cache and b) be in some utility method.

(Also, I prefer ``with`` statements; we're relying on garbage collection here to close the file descriptor)

lamby: Another approach would be to do this once per instance in ``settings.py`` and assign to VERSION = "whatever".

lamby: It also requires some kind of fallback- lots of things clean ".git" etc. etc so falling back to "(unknown)" on any error seems sensible.

WGierke: Thanks for your feedback @lamby ! I think I changed everything accordingly. Though, when creating a new commit or switching branches, the version number is not updated. One is now required to restart the docker container but I think that shouldn't be such a big problem?

lamby:
Thanks @WGierke... merged!

PR #73, opened 2017-08-25
#9 Layout base Django template to match wireframes

WGierke:
This should add a simple and extendable Django base template to build on further UI.

## Description
I added a base template and made index.html extend it.

This should implement #9 

## Motivation and Context
We'd like to start at some point with a nice and extendable UI.

## How Has This Been Tested?
I fired up `docker-compose` and saw the following page
screencapture-localhost-8000-1503916716137

## CLA


lamby:
@WGierke How come we want to serve our own assets for bootstrap etc. (curious...)

WGierke:
@lamby This way we are independent of third-party servers and can be sure that Bootstrap/JQuery/... are really always served (with the specific version we expect) and you can also develop locally.
Though, when checking out [this SO thread](https://stackoverflow.com/questions/3832446/benefits-vs-pitfalls-of-hosting-jquery-locally) it appears to be faster including the assets from third-party servers because the browser can only open a small amount of connections at the same time to the same domain.
What do you think? Should I change it again so we load the assets from external servers for better performance?

yazan-alomari:
@WGierke Yes serving assets locally is better than using a CDN in this case. We can't assume computers at clinics and hospitals have an internet connection. Usually there will be a local network with zero or limited internet connectivity.

lamby:
> Should I change it again so we load the assets from external servers for better performance?

@WGierke Nah, I think it's good. Just wanted to check there wasn't another reason why (perhaps that reasoning should have been explicit in the commit message?)

Can you add some screenshots to this PR? :)

WGierke:
@lamby Alright, sorry for the confusion. I merged the current master and added a screenshot of the current version (including the Git version in the top right corner).

interface/frontend/base.html
```diff
@@ -1,58 +1,61 @@
...
     ================================================== -->
-    <!-- Placed at the end of the document so the pages load faster -->
-    <script src="../static/js/jquery-3.1.1.slim.min.js"></script>
-    <script src="../static/js/tether.min.js"></script>
-    <script src="../static/js/bootstrap.min.js"></script>
-    <!-- IE10 viewport hack for Surface/desktop Windows 8 bug -->
-    <script src="../static/js/ie10-viewport-bug-workaround.js"></script>
-  </body>
+  <!-- Placed at the end of the document so the pages load faster -->
+  <script src="../static/js/jquery-3.1.1.slim.min.js"></script>
```
lamby: Sorry for not noticing this before, but I think we should be using Django's staticfiles framework. This ensures  — or can at least provide -- some kind of cachebusting support.

lamby:
For me, this is just blocking on moving to ``staticfiles`` I think.

interface/frontend/base.html
WGierke: @lamby you're right, thanks for the suggestion. I updated everything accordingly.

lamby:
Thanks @WGierke 

@isms Before I merge this, can you confirm that we are doing the ``collectstatic`` dance?

isms:
@lamby No `collectstatic` yet, as everyone is (probably) using the dev server: https://github.com/concept-to-clinic/concept-to-clinic/blob/master/compose/interface/start-dev.sh

Not a blocker at the moment in my mind; "production" is still a hazy concept at the moment which more than anything else currently means "not `DEBUG` mode".

Does that answer the question?

lamby:
@isms wrote:

> No collectstatic yet
[..]
> Does that answer the question?

Yep, thanks!

Issue #39, opened 2017-08-04
Inspect a DICOM directory and create appropriate model instances


isms:
## Overview
The interface backend has some model stubs for tracking metadata about DICOM imagery. The `Case` [model](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/interface/backend/cases/models.py) tracks a radiologist's workflow examining an image series, so it's the heart of this application. But the `Case` has foreign keys to the `ImageSeries` [model](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/interface/backend/images/models.py), which needs to know about the data.

## Expected Behavior
We should be able to pass a directory URI to this method (helpful ref [here](https://stackoverflow.com/questions/5977576/is-there-a-convenient-way-to-map-a-file-uri-to-os-path)) and expect that an `ImageSeries` object should be created if necessary, with metadata fields filled in.

## Technical details
- The [DICOM standard](http://dicom.nema.org/standard.html) may be helpful here
- The [pydicom](http://pydicom.readthedocs.io/en/stable/getting_started.html) library is excellent and probably what we want to use
- This should likely live in the `ImageSeries` class, possibly as a `classmethod`; this could be like:
   ```python
   uri = 'file:///path/to/project/tests/assets/LIDC-IDRI-0001/1.3.6.1.4.1.14519.5.2.1.6279.6001.298806137288633453246975630178/1.3.6.1.4.1.14519.5.2.1.6279.6001.179049373636438705059720603192'
   ImageSeries.get_or_create(uri)
   ```

## Out of scope
- Adding new fields to the models (let's get the ones we have working first)
- Using or storing the array data; an issue already exists to pull in the DICOM image's numerical array (#12) for the prediction service, but that doesn't have anything to do with our DB models on the interface side of the house.

## Acceptance criteria
- [ ] Working implementation of the method described above
- [ ] Tests added

*NOTE: All PRs must follow the standard PR checklist.*

PR #74, opened 2017-08-25
#39 Inspect a DICOM directory and create appropriate model instances

WGierke:
Get or create an `ImageSeries` instance given the path to a DICOM directory.

## Description
I added the class function `get_or_create` to the `ImageSeries` class.

This should solve #39 

## Motivation and Context
The `ImageSeries` model needs to know the data it's holding.

## How Has This Been Tested?
I wrote a test for it.

## CLA

### DIFF
--- a/interface/backend/images/models.py
+++ b/interface/backend/images/models.py
@@ -1,4 +1,6 @@
 from django.db import models
+import os
+import dicom
 
 
 class ImageSeries(models.Model):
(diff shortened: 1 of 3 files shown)


interface/backend/images/models.py
```diff
@@ -11,6 +13,22 @@ class ImageSeries(models.Model):
...
+    def get_or_create(uri):
+        """
+        Given the absolute uri to a directory with DICOM images of a patient,
+        look up the ImageSeries with the same PatientID and SeriesInstanceUID.
+        If none exists so far, create one.
+        Return a tuple of (ImageSeries, created), where created is a boolean specifying whether the object was created.
+        :return: (ImageSeries, bool)
+        """
+        file_ = glob.glob1(uri, '*.dcm')[0]
+        plan = dicom.read_file(uri + "/" + file_)
```
lamby: ``os.path.join``? Or even Django's ``safe_join`` rather..  Oherwise we could have some fun with ``../../../../lol`` I fear

WGierke: Very good point, thank you @lamby ! I updated it with `safe_join`.

interface/backend/images/models.py
```diff
@@ -23,3 +47,4 @@ class ImageLocation(models.Model):
     y = models.PositiveSmallIntegerField(help_text='Voxel index for Y axis, zero-index, from top left')
 
     z = models.PositiveSmallIntegerField(help_text='Slice index for Z axis, zero-index')
+
```
lamby: Please be careful of checking your commits before pushing... You've read about Van Halen's brown M&M's ider, right? :)

lamby:
thanks @WGierke !

Issue #13, opened 2017-08-04
Calculate tumor volume from pixel masks


pjbull:
## Overview
We want to provide clinicians with summary statistics about identified tumors. In the `segment` model, we output per-pixel binary masks that identify which pixels are likely to be cancer. From these masks, we want to calculate the volume of the tumor. 

## Expected Behavior
Take as an input the path to the serialized masks (which is the output of the segment `predict` method). We also have the centroids that are passed into the `predict` method. For each centroid, calculate the volume of the tumor. DICOM has slice sizes so the units should be in real measurements (not pixels).

Design doc reference:
[Jobs to be done > Segment](https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#segment)

## Technical details
 - This feature should be implemented in the `trained_model.py` file.
 - The output should be the volume for each tumor (from the centroid list) in a real unit

## Related issues

#XX is the work to add this output to the API endpoint.

## Acceptance criteria
 - [ ] method that outputs volumes from a binary mask and list of centroids
 - [ ] tests for this method

*NOTE: All PRs must follow the standard PR checklist.*

vessemer:
I'm a bit confused: how we are going to calculate the real volume of an area while the function does not receive the CT meta data. Also, can you please clarify how the output from the predict method will be serialised? And last: is it necessary to provide the centroids, since all centroids may be re-estimated via the connected components analysis?

reubano:
Hi @vessemer

> I'm a bit confused: how we are going to calculate the real volume of an area while the function does not receive the CT meta data. 

Good point! I think one way to approach this is to create a new function that outputs `metadata`, and then pass in that `metadata` as a third argument. I see you've assumed something similar in your PR via `voxel_shape`.

> Also, can you please clarify how the output from the predict method will be serialised? 

That will be up to whoever implements the `predict` function. I see you have assumed `npy` files. That should be perfectly fine as a stub for now. In the event a different serialization method is chosen, it shouldn't be too difficult to make the appropriate updates.

> And last: is it necessary to provide the centroids, since all centroids may be re-estimated via the connected components analysis?

[...]

PR #78, opened 2017-08-27
Calculate tumor volume

vessemer:
## Description
This method computes the volume of connected components placed at the specified centroids, via connected component analysis using `scipy.ndimage.label`. The `voxel_shape` parameter has added to the function in order to derive units of measure. 

Calculate tumor volume from pixel masks [#13](https://github.com/concept-to-clinic/concept-to-clinic/issues/13) 

## How Has This Been Tested?
The unit test for `calculate_volume` has been added to test for the volume calculation over the different mask instances. During the testing, another issue has acquired at `test_segment`, which was related to the meaningless `segment_path`. Hence the changes were also added to the `segment.predict` function.

## CLA

### DIFF
--- a/prediction/src/algorithms/segment/trained_model.py
+++ b/prediction/src/algorithms/segment/trained_model.py
@@ -7,7 +7,11 @@
     descriptive statistics.
 """
 
-from src.preprocess.load_dicom import load_dicom
+from src.preprocess.load_dicom import load_dicom, load_meta
+
+import numpy as np
+import os
+import scipy.ndimage
 
 
 def predict(dicom_path, centroids):
@@ -35,27 +39,52 @@ def predict(dicom_path, centroids):
              'volumes': list[float]}
     """
     load_dicom(dicom_path)
-    segment_path = 'path/to/segmentation'
+    segment_path = os.path.join(os.path.dirname(__file__),
+                                'assets', 'test_mask.npy')
     volumes = calculate_volume(segment_path, centroids)
     return_value = {
         'binary_mask_path': segment_path,
         'volumes': volumes
     }
+
     return return_value
 
 
-def calculate_volume(segment_path, centroids):
-    """ Calculates tumor volume from pixel masks
+def calculate_volume(segment_path, centroids, dicom_path=None):
+    """ Calculates tumor volume in cubic mm if a dicom_path has been provided.
 
+    Given the path to the serialized mask and a list of centroids
+        (1) For each centroid, calculate the volume of the tumor.
+        (2) DICOM has voxels' sizes in mm therefore the volume should be in real
+        measurements (not pixels).
     Args:
-        segment_path (str): A path to the serialized binary mask for
-            each centroid
+        segment_path (str): a path to a mask file
         centroids (list[dict]): A list of centroids of the form::
             {'x': int,
              'y': int,
              'z': int}
+        dicom_path (str): contains the path to the folder containing the dcm-files of a series.
+            If None then volume will be returned in voxels.
 
     Returns:
-        list[float]: List of volumes per centroid
+        list[float]: a list of volumes in cubic mm (if a dicom_path has been provided)
+            of a connected component for each centroid.
     """
-    return [0.5 for centroid in centroids]
+
+    mask = np.load(segment_path)
+    mask, _ = scipy.ndimage.label(mask)
+    labels = [mask[centroid['x'], centroid['y'], centroid['z']] for centroid in centroids]
+    volumes = np.bincount(mask.flatten())
+    volumes = volumes[labels].tolist()
+
+    if dicom_path:
+        dicom_files = load_meta(dicom_path)
+        slice_locations = [dcm_file.SliceLocation for dcm_file in dicom_files]
+        slice_thickness = np.diff(slice_locations).mean()
+        # Every DICOM study preserve the same PixelSpacing along its sub files
+        voxel_shape = dicom_files[0].PixelSpacing
+        # Taking into account ijk -> xyz transformation
+        voxel_shape = np.prod([voxel_shape[1], voxel_shape[0], slice_thickness])
+        volumes = [volume * voxel_shape for volume in volumes]
+
+    return volumes
--- a/prediction/src/tests/test_calculate_volume.py
+++ b/prediction/src/tests/test_calculate_volume.py
@@ -1,0 +1,69 @@
+import numpy as np
+import pytest
+
+from ..algorithms.segment import trained_model
+
+
+def generate_motes(mask, centroid, volume):
+    centroid_ = np.asarray([centroid['x'], centroid['y'], centroid['z']])
+    free_voxels = np.where(mask != -1)
+    free_voxels = np.asarray(free_voxels).T
+    free_voxels = sorted(free_voxels, key=lambda x: np.linalg.norm(x - centroid_, ord=2))
+    free_voxels = np.asarray(free_voxels[:volume]).T
+    mask[(free_voxels[0], free_voxels[1], free_voxels[2])] = True
+
+
+def generate_mask(shape, centroids, volumes):
+    mask = np.zeros(shape, dtype=np.bool_)
+    for centroid, volume in zip(centroids, volumes):
+        generate_motes(mask, centroid, volume)
+    return mask
+
+
+def test_calculate_volume_over_unconnected_components(tmpdir):
+    centroids = [[0, 0, 0], [32, 32, 28], [45, 45, 12]]
+    centroids = [{'x': centroid[0], 'y': centroid[1], 'z': centroid[2]} for centroid in centroids]
+    mask = generate_mask(shape=[50, 50, 29], centroids=centroids, volumes=[100, 20, 30])
+    # The balls modelled to be not overlapped
+    assert mask.sum() == 150
+
+    path = tmpdir.mkdir("masks").join("mask.npy")
+    np.save(str(path), mask)
+
+    volumes_calculated = trained_model.calculate_volume(str(path), centroids)
+    assert len(volumes_calculated) == 3
+    assert volumes_calculated == [100, 20, 30]
+
+
+@pytest.fixture(scope='session')
+def get_mask_connected(tmpdir_factory):
+    centroids = [[0, 0, 0], [0, 0, 0], [45, 45, 12]]
+    centroids = [{'x': centroid[0], 'y': centroid[1], 'z': centroid[2]} for centroid in centroids]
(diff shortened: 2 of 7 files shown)


WGierke:
I'd suggest you to use `flake8` here as well :)
```
➜  concept-to-clinic git:(13_calculate_tumor_volume) flake8 prediction
prediction/src/algorithms/segment/trained_model.py:53:66: W291 trailing whitespace
prediction/src/algorithms/segment/trained_model.py:54:81: W291 trailing whitespace
prediction/src/algorithms/segment/trained_model.py:62:77: W291 trailing whitespace
prediction/src/algorithms/segment/trained_model.py:63:65: W291 trailing whitespace
prediction/src/algorithms/segment/trained_model.py:65:12: W291 trailing whitespace
prediction/src/algorithms/segment/trained_model.py:67:1: W293 blank line contains whitespace
prediction/src/algorithms/segment/trained_model.py:76:1: W293 blank line contains whitespace
prediction/src/tests/test_calculate_volume.py:17:1: W293 blank line contains whitespace
prediction/src/tests/test_calculate_volume.py:24:1: W293 blank line contains whitespace
prediction/src/tests/test_calculate_volume.py:35:48: W291 trailing whitespace
prediction/src/tests/test_calculate_volume.py:38:1: W293 blank line contains whitespace
prediction/src/tests/test_calculate_volume.py:41:47: W291 trailing whitespace
prediction/src/tests/test_calculate_volume.py:43:1: W293 blank line contains whitespace
prediction/src/tests/test_calculate_volume.py:46:1: E303 too many blank lines (3)
prediction/src/tests/test_calculate_volume.py:52:1: W293 blank line contains whitespace
prediction/src/tests/test_calculate_volume.py:59:1: W293 blank line contains whitespace
prediction/src/tests/test_calculate_volume.py:64:1: W293 blank line contains whitespace
```

vessemer:
Thanks, @WGierke, I've followed your suggestion.

prediction/src/tests/test_calculate_volume.py
```diff
@@ -0,0 +1,68 @@
...
+    return mask
+
+
+def test_calculate_volume_on_unoverlapped_connected_components():
+    centroids = [[0, 0, 0], [32, 32, 28], [45, 45, 12]]
+    centroids = [{'x': centroid[0], 'y': centroid[1], 'z': centroid[2]} for centroid in centroids]
+    mask = generate_mask(shape=[50, 50, 29], centroids=centroids, volumes=[100, 20, 30])
+    assert mask.sum() == 150
+
+    path = os.path.join('.', 'tmp', 'mask.npy')
```
lamby: Please use pylint's ``tmpdir`` support; it will — at the very least — clean up after itself on failing tests and avoid the need for ugly path manipulation!

vessemer: Thank you for pointing it out, I've fixed it.

prediction/src/algorithms/segment/trained_model.py
```diff
@@ -32,27 +37,59 @@ def predict(dicom_path, centroids):
...
+    volumes = np.bincount(mask.flatten())
+    volumes = volumes[labels]
+
+    return volumes.tolist()
+
+
+def to_real_volume(volumes, dicom_path):
+    """
+    Transforms the volumes, provided in voxel units, into the volumes in mm.
+    The metadata the DICOM.
```
reubano: Looks like a typo here :)

reubano:
Hi @vessemer thanks for the updates! I've just been trying to wrap my head around what you've done, so hopefully you can help me understand :). 

Earlier you mentioned concern about not having access to the dicom metadata, and passed in a third volume parameter to the `calculate_volume` function to accommodate. I now see that third parameter is no longer present. Did you figure out a clever way perform the calculation without it?

Also, what is the use case for `to_real_volume`? I see you have a test for it, but it isn't referenced by the `predict` function.

vessemer:
Hi @reubano, as you requested in [#13 (comment)](https://github.com/concept-to-clinic/concept-to-clinic/issues/13#issuecomment-325324091), 

> While it would be possible for calculate_volume to take dicom_path as an argument, and then calculate centroids on its own, we prefer to uncouple those steps and have a canonical function dedicated to that task.

I've decoupled the functions so that now the `calculate_volume` returns the volume in voxels, and only `to_real_volume` takes into account the voxels' shape and transform the list of volumes in voxels into the list of volumes in mm.

reubano:
Sorry if I wasn't clear. By "uncouple those steps", I was referring to not calculating `centroids` from a `dicom_path` inside the `calculate_volume` function. So, the upshot of that was that `calculate_volume` would just accept a `centroids` parameter (which was calculated by the decoupled [`identify/trained_model/predict`](https://github.com/concept-to-clinic/concept-to-clinic/blob/5ceab8915d96268d34f638abbcf4b71981cc6d6d/prediction/src/algorithms/identify/trained_model.py#L13) function).

reubano:
So feel free to re-integrate `to_real_volume` unless you see an advantage for keeping it separate.

reubano:
Hi @vessemer, just wanted to check and make sure everything is ok. Is there anything blocking you? I think if you address the `to_real_volume` function (either by incorporating it into the `predict` function, or indicating when and how it should be used) when we can move forward to getting this merged. Please let me know if anything is unclear or if you have any questions.

vessemer:
@reubano I've addressed the `to_real_volume` by incorporating it into the `calculate_volume` function.  I've also added the function `load_meta` into the `load_dicom.py` file, in order to uncouple ugly `os.path.join(dicom_path, '*.dcm')` from the `calculate_volume`. It's worth noting that there exists not only DICOM files, but also the .mhd/.raw which has been used e.g. in LUNA16. Thus it may be useful to support different formats as well.

prediction/src/tests/test_calculate_volume.py
```diff
@@ -0,0 +1,71 @@
...
+    print(path, centroids)
+    volumes_calculated = trained_model.calculate_volume(str(path), centroids)
+    #   Despite they are overlapped, the amount of volumes must have preserved
+    assert len(volumes_calculated) == 3
+    assert volumes_calculated == [100, 100, 30]
+
+
+def test_calculate_volume_over_connected_components_with_dicom_path(get_mask_connected):
+    path, centroids = get_mask_connected
+    print(path, centroids)
```
reubano: Was this `print` meant to be left in? Perhaps comment it out, or use an `assert` statement if the result of the printing is necessary?

prediction/src/tests/test_calculate_volume.py
```diff
@@ -0,0 +1,71 @@
...
+    assert mask.sum() == 130
+
+    path = tmpdir_factory.mktemp("masks").join("mask.npy")
+    np.save(str(path), mask)
+    return path, centroids
+
+
+def test_calculate_volume_over_connected_components(get_mask_connected):
+    path, centroids = get_mask_connected
+    print(path, centroids)
```
reubano: Was this `print` meant to be left in? Perhaps comment it out, or use an `assert` statement if the result of the printing is necessary?

prediction/src/tests/test_calculate_volume.py
```diff
@@ -0,0 +1,71 @@
...
+    voxels_volumes = [100, 100, 30]
+    dicom_path = '../images/LIDC-IDRI-0001/1.3.6.1.4.1.14519.5.2.1.6279.6001.298806137288633453246975630178/' \
+                 '1.3.6.1.4.1.14519.5.2.1.6279.6001.179049373636438705059720603192'
+    real_volumes = trained_model.calculate_volume(str(path), centroids, dicom_path)
+
+    #   Despite they are overlapped, the amount of volumes must have preserved
+    assert len(real_volumes) == len(voxels_volumes)
+    assert all([(mm <= 1.236 * vox) or (mm >= 1.234 * vox)
+                for vox, mm in zip(voxels_volumes, real_volumes)])
```
reubano: That's a pretty large range :). Any chance this should be an `and` instead of an `or` and the directions reversed? If so, we could also simplify it like so `1.234 >= mm / vox >= 1.236`.

prediction/src/algorithms/segment/trained_model.py
```diff
@@ -35,27 +39,53 @@ def predict(dicom_path, centroids):
...
+    if dicom_path is None:
+        return volumes.tolist()
+
+    dicom_files = load_meta(dicom_path)
+    slice_locations = [dcm_file.SliceLocation for dcm_file in dicom_files]
+    slice_thikness = np.diff(slice_locations).mean()
+    # Every DICOM study preserve the same PixelSpacing along its sub files
+    voxel_shape = dicom_files[0].PixelSpacing
+    # Taking into account ijk -> xyz transformation
+    voxel_shape = np.prod([voxel_shape[1], voxel_shape[0], slice_thikness])
```
reubano: `slice_thikness`--> `slice_thickness` :)

prediction/src/algorithms/segment/trained_model.py
```diff
@@ -35,27 +39,53 @@ def predict(dicom_path, centroids):
...
-    return [0.5 for centroid in centroids]
+
+    mask = np.load(segment_path)
+    mask, _ = scipy.ndimage.label(mask)
+    labels = [mask[centroid['x'], centroid['y'], centroid['z']] for centroid in centroids]
+    volumes = np.bincount(mask.flatten())
+    volumes = volumes[labels]
+
+    if dicom_path is None:
+        return volumes.tolist()
```
reubano: So this will still let empty paths go through, and two `return` statements aren't really ideal. Would you be able to refactor into something like this?

```python 
if dicom_path:
    pass
else:
    pass

return volumes
```

reubano:
> It's worth noting that there exists not only DICOM files, but also the .mhd/.raw which has been used e.g. in LUNA16. Thus it may be useful to support different formats as well.

Nice observation! Would you like to submit an issue for this use case?

prediction/src/tests/test_calculate_volume.py
vessemer: Done, just >= and <= are swaped.

vessemer:
> Nice observation! Would you like to submit an issue for this use case?  

I've created [one](https://github.com/concept-to-clinic/concept-to-clinic/issues/98), but I also guess that it'll be necessary to implement the class aimed at metadata standardisation, maybe along with all reading orchestration.

reubano:
LGTM. This just needs to be squashed to one commit and it'll be ready to merge!

reubano:
Hi @vessemer, please let me know if you need any assistance squashing the commits. You should be able to do something like `git rebase -i master`.

reubano:
Awesome @vessemer! :shipit:

Issue #18, opened 2017-08-04
Document the grt123 algorithm


reubano:
## Overview

Participants in the [Data Science Bowl](https://www.kaggle.com/c/data-science-bowl-2017) produced several algorithms that we would like to incorporate. To help facilitate this effort, we also want to add documentation so that contributors can make an educated decision when selecting an algorithm to incorporate.

## Expected Behavior

This documentation should enable people to:
 - view the library dependencies and license
 - understand its pros/cons
 - evaluate its performance/accuracy
 - identify which areas of the codebase to target for improvement

Design doc reference: [Detect and select](https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#detect-and-select)

## Algorithm info

key | value
----|------
team|grt123
rank|1
repo|https://github.com/lfz/DSB2017
trained models|https://github.com/lfz/DSB2017/tree/master/model
converted branch|https://github.com/concept-to-clinic/DSB2017
ML engine|pytorch
engine-version|0.1.10+ac9245a
ML backend|
backend-version|
training method|
architecture|
algorithm|
OS|Ubuntu
OS version|14.04
Python version|2.7
CUDA version|8
cuDNN version|5.1
notes|

## Technical details

 - This feature should be implemented as a new markdown file in the [`docs folder`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/docs)

## Acceptance criteria

 - [ ] effective documentation for the above

*NOTE: All PRs must follow the standard PR checklist.*

PR #82, opened 2017-08-29
#18 Document the grt123 algorithm

WGierke:
I added an initial documentation of the grt123 algorithm. As soon as I have some specs on how long the prediction takes, I'll replace the TODOs. Please feel free to add commits and extend it!

This pull request addresses #18 .

## Motivation and Context

## CLA

### DIFF
--- a/docs/algorithm_grt123.md
+++ b/docs/algorithm_grt123.md
@@ -1,0 +1,146 @@
+# grt123 Algorithm
+
+## Summary
+The algorithm consists of nodule detection and cancer classification based on 3D convolution neural networks.
+The model receives preprocessed 3D lung scans as
+input and outputs both the bounding boxes of suspicious nodules and the probability of getting cancer.
+
+## Source
+**Author:** Team grt123      
+**Repository:** https://github.com/lfz/DSB2017  
+1st place at the Data Science Bowl 2017
+
+## License
+Neither mentioned in the repository nor in the technical report, but since the authors had to accept the [Data Science Bowl rules](https://www.kaggle.com/c/data-science-bowl-2017/rules), the code must be published under [MIT](http://opensource.org/licenses/MIT) license.
+
+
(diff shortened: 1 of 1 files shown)


reubano:
Nice writeup! My only concern is the ability to run on a CPU. In familiarizing myself with the DSB, I [attempted to port](https://github.com/concept-to-clinic/DSB2017) this algorithm to Python 3. In the process I encountered several hurdles when also attempting to run on a CPU. You can see some of the changes I made [such as this](https://github.com/concept-to-clinic/DSB2017/blob/877622c3c18fb48c8004e0afb8472f916a9d3277/predict/__init__.py#L52) in the [conversion branch](https://github.com/concept-to-clinic/DSB2017/tree/conversion).

WGierke:
Thanks for your feedback! I'm not familiar with pyTorch at all but I would have never thought that it's so complicated to port it to Python 3 enable it to run on a CPU... In that case I should clearly update the relevant parts of the algorithm documentation about that. DId you happen to try how long it takes the model to predict the cancer probability on one image?

reubano:
> Did you happen to try how long it takes the model to predict the cancer probability on one image?

I ran the prediction algo on the sample image set in a AWS p2 instance. I don't rem the timings exactly. Maybe a few mins. Didn't do so for a single image since the repo wasn't set up to work that way.

reubano:
Hi @WGierke any luck updating the CPU portion? In particular, the section that says `It already is possible to use the model on both CPU and GPU.` should be modified to address the issues that may be encountered when porting to run on a CPU (such as `DataParallel`).

reubano:
:shipit:

WGierke:
Hey @reubano unfortunately I had no success so far, sorry. Nevertheless, I updated the relevant sections based on your remarks.

reubano:
No problem!

Issue #32, opened 2017-08-04
Breadcrumb navigation


lamby:
## Overview

A "breadcrumb trail" is a type of secondary navigation scheme that reveals the user's location in a website or Web application. For example, "Open image -> Detect and select -> ...". We should add this for mnavigation based on the current stage in the analysis to aid the end-user

## Expected Behavior

The current page should be highlighted:

![](https://i.imgur.com/KNyNv2D.jpg)

The links do not need to be actual hyperlinks, merely text.

## Technical details

We should avoid DRY (Don't Repeat Yourself) violations by duplicating the HTML on every possible page with slightly different markup. The links do not need to be clickable.

## Blocked by
Issues: #9

PR #84, opened 2017-08-31
#32 Breadcrumb Navigation

WGierke:
I implemented a basic DRY breadcrumb navigation. It's currently blocked by #73 as it's already using the basic bootstrap template. Since I don't have that much experience in front-end programming and Django, I'd be very happy about any feedback or commits.

## Description
I moved the navigation to [frontend/shared/tabs.html](https://github.com/concept-to-clinic/concept-to-clinic/compare/master...WGierke:32_breadcrumb_navigation?expand=1#diff-e6833f2ee1dbb5169558392ed92ee414). Every page includes that shared template by passing the name of the current page as `active_tab`:
```
{% block tabs %}
    {% include "shared/tabs.html" with active_tab='detect_and_select' %}
{% endblock %}
```
The navigation will then be rendered and the tab that indicates the current page should be highlighted:
```
...
<li class="nav-item {% if active_tab == 'detect_and_select' %}active{% endif %}">
    <a class="nav-link" href="/detect">Detect and Select</a>
</li>
...
```

This issue addresses #32 .

## Motivation and Context
The user should be able to see in the navigation on which page he currently is.

## How Has This Been Tested?
I added placeholder pages and verified that when visiting every page, each tab that belongs to the page is highlighted.

screenshot from 2017-08-31 11-44-54
screenshot from 2017-08-31 11-45-06
screenshot from 2017-08-31 11-45-20
screenshot from 2017-08-31 11-45-31

## CLA


interface/backend/static/views.py
```diff
@@ -2,5 +2,21 @@
 from django.shortcuts import render
 
 
-def home(request):
-    return render(request, '../frontend/index.html', {'short_hash': settings.APP_VERSION_NUMBER})
+def get_version_context():
```
lamby: This should be moved to a Django "context processor" so that we don't have to include it in every single request.

In fact, what I suggest you do is add a ``settings_context`` processor that injects the ``django.conf.settings`` variable. This will mean we just have to do ``{{ settings.APP_VERSION_NUMBER}}`` in our template.

interface/backend/static/views.py
```diff
@@ -2,5 +2,21 @@
...
+    return {'short_hash': settings.APP_VERSION_NUMBER}
+
+
+def open_image(request):
+    return render(request, '../frontend/open_image.html', get_version_context())
```
lamby: Hm, ``..`` in the URL looks very iffy/suspicious to me! Are our ``DIRS`` setup correctly?

interface/backend/static/views.py
WGierke: That definitely makes sense, thanks! Are there any opinions on favoring libraries over self-implemented stuff? I mean there is [django-useful](http://django-useful.readthedocs.io/en/latest/usage.html#context-processors), which offers `useful.context_processors.settings` and which does exactly what we want - but one could also [build that on your own](https://stackoverflow.com/a/433209) which adds some more hoops but no further dependency. What would you favor?

interface/backend/static/views.py
WGierke: Yes they are. I replaced everything with the simple file names of the templates and it works :)

interface/backend/static/views.py
isms: I feel strongly that this should not be tackled. Despite the validity at present of getting the layout mocked out while we wait for Vue.js usage to take over, the use of Django templates should be aggressively minimized.

Providing the short hash or something like it should be done via API, and probably won't even be its own view but rather as a piece of metadata in some general request(s) if not every response.

WGierke: @lamby Now that the initial bootstrap template has been merged, I squashed the commits together with the commits to update the template paths and to add a context processor.
@isms Does it mean that this issue/PR is completely blocked because we're waiting for the Vue.js stuff? Do you have any suggestions which call should deliver the version?

interface/backend/static/views.py
isms: > @isms Does it mean that this issue/PR is completely blocked because we're waiting for the Vue.js stuff? Do you have any suggestions which call should deliver the version?

Not at all, I'm in favor of getting these mock ups accepted as a good HTML starting point! I just don't want to have commits spinning their wheels on, e.g., backend view context processor shenanigans that are destined to get ripped right back out as the API+Vue take over from anything dynamic in Django templates.

interface/backend/static/context_processors.py
```diff
@@ -0,0 +1,5 @@
+from django.conf import settings
+
+
+def git_version(request):
+    return {'APP_VERSION_NUMBER': settings.APP_VERSION_NUMBER}
```
lamby: Did you see my idea about just including ``django.conf.settings`` in a ``settings_context`` ? That prevents us having to add other context processors in the future :)

lamby: Did you see my idea about just including ``django.conf.settings`` in a ``settings_context`` ? That prevents us having to add other context processors in the future :)

interface/frontend/annotate_and_segment.html
```diff
@@ -0,0 +1,21 @@
+{% extends 'base.html' %}
+{% load static %}
+
+{% block tabs %}
```
lamby: Whilst it makes no difference, I always do

```
{% block whatever %}
{{ block.super }}
content
{% endblock %}
```

Feels ... "safer"...

interface/frontend/shared/tabs.html
```diff
@@ -0,0 +1,34 @@
...
+    <ol class="breadcrumb"/>
+  </div>
+  <div class="col-md-12">
+    <nav class="navbar navbar-toggleable-md navbar-inverse bg-inverse fixed-top">
+      <img src="{% static 'images/c-logo-light.svg' %}"/>
+      <a class="navbar-brand" href="#">Concept To Clinic</a>
+      <div class="navbar-collapse collapse" id="navbar">
+        <ul class="navbar-nav justify-content-center">
+          <li class="nav-item {% if active_tab == 'open_image' %}active{% endif %}">
+            <a class="nav-link" href="/">Open image</a>
```
lamby: These should use the ``{% url … %}`` templatetag, no?

interface/frontend/annotate_and_segment.html
WGierke: Thanks again for that very valuable input! Now, the whole `settings` module can be accessed from the templates and the correct `url` template tags are used.

Issue #31, opened 2017-08-04
Show directory tree of possible images


lamby:
## Overview

The first step of the analysis and identification process is to select an image file from the local disk. We should therefore show the user all the possible files available.

## Expected Behavior

Upon loading the application, we should show a directory tree of the potential files:

![](https://i.imgur.com/nPNtluH.jpg)

## Technical details

The files should be displayed in a tree, ideally, using the Django ``storage`` framework. In the development environement this should be a directory within our project but ignored via a suitable ``.gitignore`` file.

No file format or name filtering is required; we can assume all files in the specified directory are valid. Sorting may need to be applied to the result of Django's ``listdir`` for deterministic ordering.

## Acceptance criteria

All files are listed in their correct hierarchy, sorted alphabetically.

## Blocked by
Issues: #9

rahit:
Hello!
Current API call in index page retrieving image list from `/api/images/`which is using magic method of Django REST frameworks's ModelViewSet to list down all `ImageSeries` object. Since `ImageSeries` is a database model, at first the files needed to be loaded into database to instantiate that way. 

On the other hand, to load files directly from filesystem, we should use Django's `storage` class as you mentioned above. I would like to know which would be the desired system flow? Or we can to use different API endpoint to retrieve files from storage. 

It would be helpful if anybody can give some clarification!

lamby:
@rahit 

Thanks so much for your question. In addition, my sincere apologies for the delay in getting back to you; this notification it stlipped through my "net" so I didn't see it until now..

Ideally, we would want to browse using the ``storage`` framework and — after selection — ``PUT`` the selected files to the API to *create* the ``ImageSeries`` at that point. This ensures a proper separation of the API and this particular frontend. Putting it another way, in this scenario, this ticket is merely about the filesystem browsing step.

However, if we need the ``ImageSeries`` to generate/show the previews, then we will assume that the ``ImageSeries`` already exists. Can you investigate this as a first step? :)

rahit:
Thank you for your response @lamby. According to current set up, it seems the system is assuming that image files would be preloaded in the database. Current vue frontend is expecting `ImageSeries`. API call is also made to an endpoint which expected to return all `ImageSeries` objects. If we want to import image information based on user's selection, I think the better option would be using `ViewSet` instead of `ModelViewSet`. In extended `ViewSet`, we would be able to decouple filesystem API from image series information API.

lamby:
@rahit 

[...]

PR #81, opened 2017-08-27
New API endpoint to fetch dataset directory

rahit:
New API for direcotory listing in homepage as described in #77 

## Description
Introducing new API endpoint which will be used to retrieve images from data directory.

## Reference to official issue
See issue #77 

## How Has This Been Tested?
This modification will effect homepage initial dataload. Implementation of the API is yet to be made.

## Updates of Acceptance Criteria
- [x] Update routes
- [x] Declare view (Not implementation)
- [x] Update index.html API call

## CLA
- [x] I have signed the CLA; if other committers are in the commit history, they have signed the CLA as well

PR #85, opened 2017-09-01
#31 Show directory tree of possible images

WGierke:
A file tree is rendered using Vue.js that should show the images the user can select from the local disk.

## Description
I based this PR on rahit's work in #81 that adds a new endpoint to fetch images from. The changes render a file tree at the index page that visualizes the response from that new endpoint, assuming the endpoint would return the files in a certain structure.

This should address #31 .

## Motivation and Context
It already adds some functionality in the front-end to render the structure of the available images. This is already done in Vue.js instead of using the Django template language. 

## How Has This Been Tested?
`ImageAvailableApiView.get()` is currently not implemented, yet. So currently, the tree only has a lonely root:
screenshot from 2017-09-02 00-41-58

However, assuming the view would return the following response
```
JsonResponse({'directories': [{'name': 'test_cases, no modules',
                      'children': ['test_case_001', 'test_case_002',
                                   'test_case_003']},
                     {'name': 'test_cases, benign',
                      'children': ['test_case_004', 'test_case_005',
                                   'test_case_006']},
                     {'name': 'test_cases, cancerous',
                      'children': ['test_case_007', 'test_case_008',
                                   'test_case_009']}]})
```
the picture referenced in issue #31 would be rendered:
screenshot from 2017-09-02 00-40-31

## CLA


rahit:
Hi @WGierke . One thing that passed on my mind when I initially started working on the issue was - whether the directory structure will be fixed or it will be dynamic. 
```
+-- root
     |-- Level 1
          |-- Level 2 directory 1
                |-- case 001
                |-- case 002
          |-- Level 2 directory 2
                |-- case 001
                |-- case 002
```
Do we want to grab all the files and directories under the sub directories as well? For instances, LIDC-IDRI data does not follow this structure. Nor the test images directory. If we move to another data source (or DICOM image server which is planned for production) the structure might be different. 

But, if that is not the case, next step will be very easy. What do you think about the issue? I think @lamby  can help us by guiding on this. :)

interface/frontend/open_image.html
```diff
@@ -99,6 +132,16 @@ <h3 class="card-title">{{ selected.patient_id }}</h3>
       selectSeries: function(series) {
         console.log(series.uri);
         this.selected = series;
+      },
+      fetchAvailableImages: function() {
+        var vm = this;
+        this.$http.get('/api/images/available').then(
```
lamby: ``{% url … %}`` here? :)

interface/frontend/open_image.html
```diff
@@ -99,6 +132,16 @@ <h3 class="card-title">{{ selected.patient_id }}</h3>
...
+          function(response) {
+            this.directoriesDict = response.body;
+          },
+          function() {
+            // error callback
```
lamby: Could you file a new issue (referencing this one) regarding investigating a global error handler? Like, some kind of drop-down that will pop up on any error so we aren't having to implement it each time...

WGierke:
Thanks @lamby , I added appropriate url tags and filed a new issue. What do you think of rahit's question - are we expecting that the API returns a directory structure deeper than two levels? That would most likely require us to render the file tree recursively. To resolve #31 so far I assumed that the tree should always look like the one in the picture of #31 so to only have two levels.

lamby:
> are we expecting that the API returns a directory structure deeper than two levels

For now, lets go with no. MVP!

WGierke:
@lamby That sounds nice :) Are there any more remarks?

lamby:
@WGierke No further comments here, am just seeing "This branch cannot be rebased due to conflicts". If you can fix that up I'll get it "live" :)

WGierke:
@lamby Thanks, I squashed the commits. Hopefully, merging should work now :)

lamby:
@WGierke Oh dear I'm seeing conflicts with the base branch now.

Also by "squash", please don't squash unrelated commits - it's always a bad smell when you see the word "and" in a commit message; feel free to split those bits out.

WGierke:
@lamby Thanks, I'll stick to that in the future. The conflicts should now be resolved.

lamby:
@WGierke Alas still seeing "This branch cannot be rebased due to conflicts" :)  ^

WGierke:
@lamby conflicts are resolved again. I'm not fooling you - after merging other pull requests into the master this branch just gets conflicts again :)

lamby:
Merged.. @WGierke Did you make that ticket re a global eror handler...?

WGierke:
@lamby yes it should be described in #86

PR #87, opened 2017-09-03
Added a note about Docker Toolbox for Windows 7 users

Serhiy-Shekhovtsov:
The Docker will not run on Windows 7 but the error is not informative and someone(like me) could spend lots of time trying to find what is the issue and how to solve it.

## Description
Windows 7 users should use **Docker Toolbox** instead of Docker. Docker installation will fail with not informative message _LightWeightInstaller has stopped working_. After enabling assembly binding logging (FusionLog) one may find, that the issue is `PresentationFramework.Aero2.dll` can't be found. But instead of trying to solve the issue, one should just use **Docker Toolbox**.

## CLA
- [ ] I have signed the CLA; if other committers are in the commit history, they have signed the CLA as well


lamby:
Thanks @Serhiy-Shekhovtsov !

Issue #35, opened 2017-08-04
Show list of candidates nodules on "Detect and select" page


lamby:
## Overview

Before more data is added to this view, we will start by simply showing the list of candidate nodules that are returned by the backend. Therefore, "Detect and select" views should show list of candidate nodules.

No accept/reject support is required here; that will be covered elsewhere.

## Expected Behavior

The list of nodule candidates should be displayed on the "Detect and select" stage of case investigation:

![](https://i.imgur.com/9zyMA9L.jpg)

## Technical details

An "accordion" interface is suggested. The predicted centroid metadata should be displayed as well, with any floating point numbers suitably truncated for human consumption.

## Acceptance criteria

The view called "Detect and select" will show the list of candidates corresponding to the current case and any relevant metadata (eg. centroids, etc.) are displayed.

## Blocked by
Issues: #9

lamby:
Fixed in #88

PR #88, opened 2017-09-03
#35 Show list of candidates nodules

WGierke:
This should add an "accordion" view that shows the candidate nodules for a selected image. The candidate fields `lidc_max_sensitiv, convnet_vgg, convnett_vgg_lidc, Slice` still need to be implemented in the back-end.

## Description
I added the front-end Vue.js implementation to display candidate nodules of an image based on the current API.

This addresses #35 .

## Motivation and Context
The doctor should be able to quickly get an overview of candidate nodules including their probability of being cancer.

## How Has This Been Tested?

I created two dummy candidates by letting the API return
```
[{		
    "centroid": {		
        "x": 54,		
        "y": 78,		
        "z": 23,		
        "series": 0		
    },		
    "probability_concerning": 0.94,		
    "case": 1		
},		
{		
   "centroid": {		
       "x": 1,		
       "y": 2,		
       "z": 3,		
       "series": 1		
    },		
    "probability_concerning": 0.94,		
    "case": 1		
}]
```
The outcome was
screenshot from 2017-09-03 01-00-39
screenshot from 2017-09-03 01-00-55

## CLA

### DIFF
--- a/interface/assets/css/project.css
+++ b/interface/assets/css/project.css
@@ -3,3 +3,7 @@
   padding: 2px;
   width: auto;
 }
+
+.top-buffer {
+  margin-top:20px;
+}
--- a/interface/backend/static/tests.py
+++ b/interface/backend/static/tests.py
@@ -4,7 +4,7 @@ from django.urls import reverse
 
 class SmokeTest(TestCase):
     def test_landing(self):
-        url = reverse('static:open_image')
+        url = reverse('static:open-image')
         resp = self.client.get(url)
         self.assertContains(resp, 'Concept to Clinic')
         self.assertEqual(resp.status_code, 200)
--- a/interface/frontend/detect_and_select.html
+++ b/interface/frontend/detect_and_select.html
@@ -3,20 +3,116 @@
 
 {% block tabs %}
 {{ block.super }}
-{% include "shared/tabs.html" with active_tab='detect_and_select' %}
+{% include "shared/tabs.html" with active_tab='detect-and-select' %}
 {% endblock %}
 
 {% block content %}
+{{ block.super }}
 <div id="app">
   <div class="row">
-    <div class="col-md-12">
-      <div class="card">
-        <div class="card-block">
-          <h3 class="card-title">Placeholder for Detect and Select</h3>
-          <p class="card-text">Placeholder for Detect and Select</p>
+    <div class="col-md-4">
+      <template v-if="candidates.length">
+        <div id="accordion" role="tablist" aria-multiselectable="true">
+        <div class="card" v-for="(candidate, index) in candidates">
+          <div class="card-header" role="tab" :id="'heading_' + index">
+            <p class="mb-0">
+              <a :class="index != 0 ? 'collapsed' : null"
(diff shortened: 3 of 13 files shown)


interface/assets/css/project.css
```diff
@@ -3,3 +3,5 @@
   padding: 2px;
   width: auto;
 }
+
+.top-buffer { margin-top:20px; }
```
lamby: I wonder if this should follow the existing style to some degree?

interface/backend/static/tests.py
```diff
@@ -4,7 +4,7 @@
 
 class SmokeTest(TestCase):
     def test_landing(self):
-        url = reverse('static:home')
+        url = reverse('static:open_image')
```
lamby: I think elsewhere we are using hyphens in url ``name`` attributes

interface/frontend/detect_and_select.html
```diff
@@ -7,16 +7,109 @@
 {% endblock %}
 
 {% block content %}
+{{ block.super }}
+{% verbatim %}
```
lamby: ``{% verbatim %}`` always seems like a bad smell to me in Django. How come we need it?

WGierke: The Vue.js template syntax collides with the Django template syntax. Django would e.g. try to render following statements like `Candidate {{ index + 1 }}` or `{{ candidate.centroid.x }}` by itself, although the variables are provided by the Vue.js template language.

interface/assets/css/project.css
WGierke: I'm not sure what you mean? Currently, the main content of a page would be overlapped at the top to some degree by the navbar. Adding some margin to the content container in `base.html` should solve that.

interface/backend/static/tests.py
WGierke: Thanks for the remark. I replaced all underscores in url names by hyphens.

interface/frontend/detect_and_select.html
lamby: > The Vue.js template syntax collides with the Django template syntax

Getcha. I was wondering if we could move the ``{% verbatim %}`` blocks to only surround the colliding javascript in that case? That would make it a bit clearer why, and implicitly so. Alternatively, we could move the JS to an external file. What do you think?

interface/assets/css/project.css
lamby: Existing CSS code/indentation/spacing style, not the final result.. sorry for the confusion!

interface/frontend/detect_and_select.html
WGierke: Yes, I'd also prefer moving the tags so that they surround the "problematic" syntax as tightly as possible.

interface/assets/css/project.css
WGierke: Oh I see! Yeah, that would be way better :)

WGierke:
@lamby Thanks again for your feedback! I implemented it and squashed the commits.

lamby:
Thanks @WGierke — lamby

Issue #36, opened 2017-08-04
Accept/reject nodule candidates interface


lamby:
## Overview

Some nodules will be identified as non-candidates by the end user. Therefore, we should provde the  ability to accept/reject nodule candidates on the "Detect and select" view.

## Expected Behavior

"Dismiss" and "Mark concerning" buttons should be persisted by the user:

![](https://i.imgur.com/wJYKqNb.jpg)

The next nodule candidate displayed until there are none left.

The user's decision should be sent back to the backend immediately upon selection for persistence, rather than aggregating the entire page's results first.

PR #93, opened 2017-09-05
#36 Accept and Reject Candidates

WGierke:
This pull request is based on #88 and also blocked by it. The buttons, that have been added in this PR, now also get a functionality (that still needs to be implemented in the back-end).

## Description
You can find the "raw" changes between PR #88 and this PR [here](https://github.com/WGierke/concept-to-clinic/compare/35_list_candidates...WGierke:36_accept_reject_candidates?expand=1) which might be easier to oversee.
I added two methods that are invoked when the Dismiss / Mark Concerning buttons are clicked. The functions receive the ID of the candidate as soon as the user triggers the buttons. The entire page's results are not aggregated.

This addresses #36 .

## Motivation and Context
The end-user should be able to specify whether candidates might be concerning or whether they can simply be dismissed.

## How Has This Been Tested?
I fixed the creation of candidates (before, there was an error stating that the serializer can't handle nested `create` calls) and created three candidates. I clicked the Dismiss and Mark buttons of candidates 1 and 3 and was ably to verify the correct API responses in the console:
screenshot from 2017-09-05 23-25-23

## CLA

### DIFF
--- a/docs/apidoc_interface/backend.static.rst
+++ b/docs/apidoc_interface/backend.static.rst
@@ -12,6 +12,14 @@ backend.static.apps module
     :undoc-members:
     :show-inheritance:
 
+backend.static.context_processors module
+----------------------------------------
+
+.. automodule:: backend.static.context_processors
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
 backend.static.tests module
 ---------------------------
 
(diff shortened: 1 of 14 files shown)


lamby:
And another with merge conflicts! Alas :(

WGierke:
And resolved :)

lamby:
@WGierke Can you rebase against master instead of merging next time? No need to introduce all these ugly merges which don't really represent anything truly meaningful. I can do this for this PR however, 1 sec...

lamby:
Thanks!

Issue #30, opened 2017-08-04
Provide a summary all of the data from a case into a single JSON report


lamby:
## Overview

Whilst initially any summary will be fairly bare, by adding a summary all of the data from a case into a single JSON report early-on in the project we can be sure that any data will/can be exported for completeness reasons, etc.

It will also likely help in debugging and development of backend components as it will avoid making manual queries to the database via SQL commands or via a Python shell such as IPython.

## Expected Behavior

We should be able to including the general notes, and details for each nodule, as well as all data about a case into a single JSON report.

## Technical details

For now, this can be a simple view that takes a ``Case`` data, generates the corresponding ``dict`` structure and pretty-prints using the ``pprint`` module within a HTML page.

The generation of the ``dict`` should be separate from the display so it can be reused and tested.

## Acceptance criteria

The summary can be downloaded and the code used to generate it has at least one simple testcase to ensure lack of obvious regressions.

PR #94, opened 2017-09-06
#30 Add JSON case reports

WGierke:
This PR is based on and blocked by #93 (because this already adds a proper creation of candidates). To see the raw changes this PR wants to add, please see [here](https://github.com/WGierke/concept-to-clinic/compare/36_accept_reject_candidates...WGierke:30_case_json_report).
Now, a JSON report can be generated for a case that includes the patient id, candidates, nodules, etc.

## Description
I fixed the creation of nodules, added the JSON report and wrote a test for that.

This addresses #30.

## Motivation and Context
This should summarize the data of a case and help debugging.

## How Has This Been Tested?
I wrote a test which adds a case with image series, candidates and nodules and which tests for a proper report.

screenshot from 2017-09-06 15-43-02

## CLA

### DIFF
--- a/interface/backend/api/serializers.py
+++ b/interface/backend/api/serializers.py
@@ -54,3 +54,10 @@ class NoduleSerializer(serializers.HyperlinkedModelSerializer):
         read_only_fields = ('created',)
 
     centroid = ImageLocationSerializer()
+
+    def create(self, validated_data):
+        return Nodule.objects.create(
+            case=validated_data['case'],
+            candidate=validated_data['candidate'],
+            centroid=ImageLocation.objects.create(**validated_data['centroid']),
+        )
--- a/interface/backend/api/urls.py
+++ b/interface/backend/api/urls.py
@@ -6,12 +6,14 @@ from backend.api.views import (
     ImageAvailableApiView,
     candidate_mark,
     candidate_dismiss,
+    case_report,
 )
 from django.conf.urls import (
     include,
     url
 )
 from rest_framework import routers
+from rest_framework.urlpatterns import format_suffix_patterns
 
 router = routers.DefaultRouter()
 router.register(r'cases', CaseViewSet)
@@ -26,3 +28,6 @@ urlpatterns = [
     url(r'^candidates/(?P<candidate_id>\d+)/dismiss$', candidate_dismiss, name='candidate-dismiss'),
     url(r'^candidates/(?P<candidate_id>\d+)/mark$', candidate_mark, name='candidate-mark'),
 ]
+
+# Support different suffixes
+urlpatterns += format_suffix_patterns([url(r'^cases/(?P<case_id>\d+)/report$', case_report, name='case-report')])
--- a/interface/backend/api/views.py
+++ b/interface/backend/api/views.py
@@ -1,3 +1,4 @@
+import json
 import os
 
 from backend.api import serializers
@@ -5,12 +6,18 @@ from backend.cases.models import (
     Case,
     Candidate,
     Nodule,
+    CaseSerializer
 )
 from backend.images.models import ImageSeries
 from django.conf import settings
 from django.core.files.storage import FileSystemStorage
+from django.core.serializers.json import DjangoJSONEncoder
+from django.shortcuts import get_object_or_404
+from rest_framework import renderers
 from rest_framework import viewsets
 from rest_framework.decorators import api_view
+from rest_framework.decorators import renderer_classes
+from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
@@ -92,3 +99,20 @@ def candidate_mark(request, candidate_id):
 @api_view(['GET'])
 def candidate_dismiss(request, candidate_id):
     return Response({'response': "Candidate {} was dismissed".format(candidate_id)})
+
+
+class JsonHtmlRenderer(renderers.BaseRenderer):
+    media_type = 'text/html'
+    format = 'html'
+
+    def render(self, data, media_type=None, renderer_context=None):
+        return "<pre>{}</pre>".format(json.dumps(data, indent=4, sort_keys=True, cls=DjangoJSONEncoder))
+
+
+@api_view(['GET'])
+# Render .json and .html requests
+@renderer_classes((JSONRenderer, JsonHtmlRenderer))
+def case_report(request, case_id, format=None):
+    case = get_object_or_404(Case, pk=case_id)
+
+    return Response(CaseSerializer(case).data)
--- a/interface/backend/cases/models.py
+++ b/interface/backend/cases/models.py
@@ -1,9 +1,11 @@
+from backend.images.models import ImageSeriesSerializer, ImageLocationSerializer
 from django.core.validators import (
     MaxValueValidator,
     MinValueValidator
 )
 from django.db import models
 from django.utils import timezone
+from rest_framework import serializers
 
 
 class Case(models.Model):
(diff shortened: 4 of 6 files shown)


lamby:
@WGierke Still conflicting; please rebase :)

lamby:
@WGierke Haha. Oh dear. And I just tried to merge them using github's new UI which didn't seem to do anything :p

interface/backend/api/serializers.py
```diff
@@ -54,3 +54,11 @@ class Meta:
         read_only_fields = ('created',)
 
     centroid = ImageLocationSerializer()
+
+    def create(self, validated_data):
+        case_data = validated_data.pop('case')
```
lamby: Why `pop` the data? I mean, versus just indexing it with `validated_data['case']` (ie. please add a comment to the code itself if there is a reason; this is non-intuitive)

interface/backend/api/views.py
```diff
@@ -55,3 +61,38 @@ def candidate_mark(request, candidate_id):
...
+    case_dict['candidates'] = candidates
+
+    nodules = [model_to_dict(nodule) for nodule in Nodule.objects.filter(case_id=case_id)]
+    for nodule in nodules:
+        nodule['centroid'] = model_to_dict(ImageLocation.objects.get(pk=nodule['centroid']))
+    case_dict['nodules'] = nodules
+    return case_dict
+
+
+def json_serial(obj):
```
lamby: Looks like something we want in a globalish `utils` file instead of here. :)

interface/backend/api/views.py
```diff
@@ -55,3 +61,38 @@ def candidate_mark(request, candidate_id):
...
+
+    if isinstance(obj, (datetime, date)):
+        return obj.isoformat()
+    raise TypeError("Type %s not serializable" % type(obj))
+
+
+def case_report(request, case_id):
+    case_dict = generate_case_dict(case_id)
+    case_pretty_json = json.dumps(case_dict, indent=4, sort_keys=True, default=json_serial)
+    return HttpResponse(case_pretty_json, content_type="application/json")
```
lamby: Hm, do we not have a `JsonResponse` ?

WGierke: I'd also prefer pure JSON put the issue states `For now, this can be a simple view that takes a Case data, generates the corresponding dict structure and pretty-prints using the pprint module within a HTML page.` Doesn't that mean the rendered page should be a HTML page including the pretty printed JSON?

interface/backend/api/serializers.py
WGierke: :man_facepalming: Yep, `pop`ing definitely makes no sense here :)

interface/backend/api/views.py
WGierke: In fact, sending `application/json` as content type makes it possible to use browser add-ons that make JSON explorable and even prettier:
screenshot from 2017-09-10 19-27-55
Without such add-ons, the JSON is still printed in a nice way due to the HttpResponse.
screenshot from 2017-09-10 19-28-09

lamby: But surely JsonResponse sends exactly that? (Or we should have our own version?)

WGierke: When I use JsonResponse and try to pretty-print the JSON using `json.dumps(case_dict, indent=4, sort_keys=True..`, I end up with 
screenshot from 2017-09-11 08-51-21
which is obviously neither pretty nor valid JSON.
Without using `json.dumps` and without using a browser add-on for pretty JSON formatting, you end up with
screenshot from 2017-09-11 08-49-56
which is also not pretty.
So I'd clearly plead for the current HttpResponse and json.dumps implementation.

isms: @WGierke (cc @lamby) Seems to me neither `HttpResponse` nor `JsonResponse` is the idiomatic approach here.

Let's do this the DRF way, yeah? Please have a look at the DRF documentation, this is a solved problem: http://www.django-rest-framework.org/tutorial/2-requests-and-responses/

interface/backend/api/views.py
```diff
@@ -55,3 +60,30 @@ def candidate_mark(request, candidate_id):
 
 def candidate_dismiss(request, candidate_id):
     return JsonResponse({'response': "Candidate {} was dismissed".format(candidate_id)})
+
+
+def generate_case_dict(case_id):
+    try:
+        case_dict = model_to_dict(Case.objects.get(pk=case_id))
+    except:
```
lamby: ``except`` is too broad an exception, surely?

interface/utils.py
```diff
@@ -0,0 +1,9 @@
+from datetime import date, datetime
+
+
+def json_serial(obj):
+    """JSON serializer for objects not serializable by default json code"""
+
+    if isinstance(obj, (datetime, date)):
```
lamby: Aaaaaaactually, isn't there one of these in Django?

interface/backend/api/views.py
WGierke: Thanks @isms I wasn't aware of that idiom.
Now one can add a suffix that states what content type should be returned:
screenshot from 2017-09-12 09-53-32
screenshot from 2017-09-12 09-53-47

interface/utils.py
WGierke: Yes, I found out that `DjangoJSONEncoder` also works here :)

isms:
To be clear about what the DRF best practice is, models should be serialized with a ModelSerializer, not a custom function. Then there are lots of ways to serve that. You also don't want to manually specify the content type in general - let DRF negotiate that so it can provide the dev interface.

docs/apidoc_interface/backend.rst
```diff
@@ -9,7 +9,6 @@ Subpackages
     backend.api
     backend.cases
     backend.images
-    backend.static
```
lamby: Not sure why this is in here? :)

interface/backend/api/urls.py
```diff
@@ -26,3 +28,6 @@
     url(r'^candidates/(?P<candidate_id>\d+)/dismiss$', candidate_dismiss, name='candidate-dismiss'),
     url(r'^candidates/(?P<candidate_id>\d+)/mark$', candidate_mark, name='candidate-mark'),
 ]
+
+# Support different suffixes
+urlpatterns.extend(format_suffix_patterns([url(r'^cases/(?P<case_id>\d+)/report$', case_report, name='case-report')]))
```
lamby: FYI the typical pattern is to use ``+=``

interface/backend/api/views.py
```diff
@@ -89,3 +97,23 @@ def candidate_mark(request, candidate_id):
...
+    def render(self, data, media_type=None, renderer_context=None):
+        return "<pre>{}</pre>".format(json.dumps(data, indent=4, sort_keys=True, cls=DjangoJSONEncoder))
+
+
+@api_view(['GET'])
+# Render .json and .html requests
+@renderer_classes((JSONRenderer, JsonHtmlRenderer))
+def case_report(request, case_id, format=None):
+    try:
+        case = Case.objects.get(pk=case_id)
```
lamby: I think you want ``get_object_or_404`` here

interface/backend/api/serializers.py
```diff
@@ -54,3 +54,11 @@ class Meta:
         read_only_fields = ('created',)
 
     centroid = ImageLocationSerializer()
+
+    def create(self, validated_data):
```
lamby: This seems... quite long-winded. Why not just:

```
def create(self, validated_data):
    return Nodule.objects.create(
        case=validated_data['case'],
        candidate=validated_data['candidate'],
        centroid=ImageLocation.objects.create(**validated_data['centroid']),
    )

It's currently quite hard to follow with all the variable assignments to sliiiiiightly different names if you know what I mean?

WGierke: Nice. Short and exactly what I tried to achieve ... Thanks!

WGierke:
@lamby Thanks again for suggesting some really nice and clean shortcuts. I implemented them and squashed the commits.

lamby:
Thanks!

Issue #6, opened 2017-08-02
Set up a basic npm vue/webpack setup


isms:
## Overview
We want to be able to use `.vue` files, ES6, and so forth.

## Expected Behavior
As a first pass, [NdagiStanley/vue-django](https://github.com/NdagiStanley/vue-django) has a lot of what we want. At a minimum, we want to be able to build [.vue files](https://vuejs.org/v2/guide/single-file-components.html) so that we can use :sparkles: ES6 :sparkles and have our project JS compiled, including `.vue` files.

Design doc reference:
- Design and architecture [section](http://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#design-goals-and-architecture), software architecture diagram

## Technical details
- There is something thinking to be done about how people should develop with this setup. One awkward aspect of the `vue-django` example is that we have two devservers (I think?) -- one from Django's `manage.py` and one from the JS setup.
- **You don't have to handle hot JS reloading in this issue.**

## Acceptance criteria
- [ ] One single command to build project `.vue` (and other JS) assets, the output of which can be served by the Django process
- [ ] Docker setup updated so npm packages get installed

*NOTE: All PRs must follow the standard PR checklist.*

excalliburbd:
> You don't have to handle hot JS reloading in this issue.

Just to clarify for the first pass we want to build the Vue project and have Django serve the static files?
It's a bit confusing since the previous point mentions two servers.

isms:
@excalliburbd check out the current template, it's already using Vue this way just as a demonstration

antonow:
I've been working on this issue, and I have a few questions.

1) I see that in the "Project Structure" section of the docs there is an asset folder that is outside of the backend and frontend directories. May I propose hosting the built assets for production inside an asset folder inside the frontend folder?

By putting the assets outside the frontend project, at least one of webpack's handy features won't work. If we want to append a hash to the compiled frontend assets, so that browsers know when the production assets have been changed and need to be re-fetched, then we probably want to delete old compiled assets so they don't pile up. In order to use webpack's `clean-webpack-plugin`, the assets folder cannot be outside of the frontend project root. See screenshot below.

<img width="989" alt="screenshot" src="https://user-images.githubusercontent.com/7776175/29801623-e11c79b8-8c2d-11e7-8f5a-e05ec2785eaa.png">

[...]

PR #97, opened 2017-09-07
Set up base webpack config and install npm dependencies in Dockerfile

antonow:
This PR adds the ability to compile the frontend into static files that can be served by the django backend. It also modifies the Dockerfile to install the frontend project's npm dependencies.

## Description
I began with the vue-django webpack template and updated it to use Vue 2 and Webpack 2. The existing views built by @WGierke's and served by django were integrated into the Vue app in order to be able to leverage the VueRouter for navigating between pages. Rendering one view which initializes a SPA will make page transitions faster than asking the backend for new templates each time a link is clicked and simplifies development because all view rendering and route handling happens in one place.

I had to modify the base image for the dev Dockerfile because I was having difficulty installing node and npm with the Python 3.6 base image. I chose the latest Ubuntu image, but Debian and others would probably work too. I am not sure if there is a need to install the npm dependencies in docker however, because the `npm run build` command which compiles the production assets is run by devs outside of docker anyway, right? Or was the intention to also have Docker compile the production assets if they didn't exist in the dev's repo? If not, I can get rid of these changes to the Dockerfile.

Regarding needing to have 2 servers for frontend development, this would be the case if we wanted to have hot-reloading, but you could also run `npm run build` and serve the compiled assets with only the django backend server running. I do think we should invest the time to get the frontend dev server with hot-reloading working since this will greatly improve the frontend development process, but I was having trouble getting it working after updating Vue and Webpack and wanted to get this PR out.

[Set up a basic npm vue/webpack setup #6](https://github.com/concept-to-clinic/concept-to-clinic/issues/6)

## Motivation and Context
We need a way to compile the frontend app and serve it from django.

## How Has This Been Tested?
I ran `docker-compose -f local.yml build` to ensure that the project's npm dependencies get installed.
I ran `npm install` and `npm run build` from the frontend directory, then `docker-compose -f local.yml up`, and navigated to localhost:8000 and saw this:

screen shot 2017-09-06 at 10 51 49 pm

## CLA


interface/requirements/local.txt
```diff
@@ -3,7 +3,7 @@
 
 coverage
 factory_boy==2.8.1
-ipython==6.1.0
+--pre ipython==6.1.0
```
WGierke: Hey @antonow cool stuff! But can you tell me why this is desirable?

WGierke:
If you remove the space in front of your 'X' in the PR description, the CLA test should pass :)
Could you adapt the current failing test, that checks whether the 'Open Image' page exists, to the new Vue.js frontend? I'm still not quite sure why the second test fails ...

lamby:
> If you remove the space in front of your 'X' in the PR description, the CLA test should pass :)

No, you need to sign the CLA on https://concepttoclinic.drivendata.org - it's not "just" a text field here!

interface/requirements/local.txt
antonow: When I changed the docker base image to Ubuntu, `pip` wasn't able to find version 6.1.0 of `ipython`. I found somebody on SO saying that pre-release versions aren't displayed by default so you need to prepend `--pre` to get them.

antonow: https://pip.pypa.io/en/stable/reference/pip_install/#install-pre

antonow: My guess is that the Python 3.6 base image had the desire version of ipython already? We may be able to go back to using that base image depending on the response to the questions in my PR description.

lamby:
Hey @antonow Just needs a rebase against master now :)

antonow:
@lamby - rebased against master again

lamby:
Many thanks!

WGierke:
Hey @antonow thanks a lot for your contribution! Do you happen to have some time to fix the tests that are currently still failing? Otherwise it might sense to open an issue for that so we can keep the tests passing on the master :)

Issue #20, opened 2017-08-04
Document the Julian de Wit algorithm


reubano:
## Overview

Participants in the [Data Science Bowl](https://www.kaggle.com/c/data-science-bowl-2017) produced several algorithms that we would like to incorporate. To help facilitate this effort, we also want to add documentation so that contributors can make an educated decision when selecting an algorithm to incorporate.

## Expected Behavior

This documentation should enable people to:
 - view the library dependencies and license
 - understand its pros/cons
 - evaluate its performance/accuracy
 - identify which areas of the codebase to target for improvement

Design doc reference: [Detect and select](https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#detect-and-select)

## Algorithm info

key | value
----|------
team|Julian de Wit
rank|2
repo|https://github.com/juliandewit/kaggle_ndsb2017
trained models|https://retinopaty.blob.core.windows.net/ndsb3/trained_models.rar
converted branch|
ML engine|Keras
engine-version|
ML backend|Tensorflow
backend-version|
training method|
architecture|
algorithm|
OS|Windows
OS version|
Python version|3.5
CUDA version|
cuDNN version|
notes|

## Technical details

 - This feature should be implemented as a new markdown file in the [`docs folder`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/docs)

## Acceptance criteria

 - [ ] effective documentation for the above

*NOTE: All PRs must follow the standard PR checklist.*

Issue #19, opened 2017-08-04
Document the Daniel Hammack algorithm


reubano:
## Overview

Participants in the [Data Science Bowl](https://www.kaggle.com/c/data-science-bowl-2017) produced several algorithms that we would like to incorporate. To help facilitate this effort, we also want to add documentation so that contributors can make an educated decision when selecting an algorithm to incorporate.

## Expected Behavior

This documentation should enable people to:
 - view the library dependencies and license
 - understand its pros/cons
 - evaluate its performance/accuracy
 - identify which areas of the codebase to target for improvement

Design doc reference: [Detect and select](https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#detect-and-select)

## Algorithm info

key | value
----|------
team|Daniel Hammack
rank|1
repo|https://github.com/dhammack/DSB2017 & https://github.com/juliandewit/kaggle_ndsb2017
trained models|https://retinopaty.blob.core.windows.net/ndsb3/trained_models.rar
converted branch|
ML engine|Keras
engine-version|
ML backend|Theano
backend-version|
training method|
architecture|
algorithm|
OS|Windows
OS version|64bit
Python version|2.7 & 3.5
CUDA version|
cuDNN version|
notes|

## Technical details

 - This feature should be implemented as a new markdown file in the [`docs folder`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/docs)

## Acceptance criteria

 - [ ] effective documentation for the above

*NOTE: All PRs must follow the standard PR checklist.*

WGierke:
I've reached out to the author since he mentions in the README that he knows how to simplify and speed-up his code without major performance loss.

WGierke:
Unfortunately, the author didn't respond so far. @reubano you already wrote that the code runs on Windows 64bit with Python 3.5. From where did you receive this information? I'm a bit confused because I found a lot `print` statements in the code that don't have parentheses which would throw errors when running with Python 3.

reubano:
@wgierke turns out this one is actually a combination of two repos and it was the 2nd one that I got the info from. Thanks for pointing that out! I've since updated the description.

WGierke:
@reubano I guess this one can be closed?

reubano:
@wgierke nice catch!

kmader:
[...]

PR #99, opened 2017-09-08
#19 Document the Daniel Hammack algorithm

WGierke:
I added an initial documentation of the Daniel Hammack algorithm. The algorithm output was ensembled with the results of the [Julian de Wit algorithm](https://github.com/concept-to-clinic/concept-to-clinic/issues/20) whose documentation should follow soon. Please feel free to add commits and extend it!

This pull request addresses #19  .

## Motivation and Context

## CLA

### DIFF
--- a/docs/algorithm_hammack.md
+++ b/docs/algorithm_hammack.md
@@ -1,0 +1,164 @@
+# Daniel Hammack Algorithm
+
+## Summary
+The algorithm stands out with the idea to pre-detect possible abnormal regions which are then fed to two ensembles of 17 3D convolutional neural networks (CNNs). It is combined with the algorithm by Julian de Wit at the prediction level.
+
+## Source
+**Author:** Daniel Hammack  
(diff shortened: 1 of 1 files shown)


timothyman:
I have been looking at the Julian de Wit model myself, and I was preparing to do a test run on my local machine -- it took a bit longer than expected since NSDB3 has a lot of training data and my harddisk is too small.

My approach was to get the model running locally, so that I am sure that the current versions of libraries and CUDA/cuDNN versions work with this model, after which we can freeze the requirements for the documentation. For instance, doing this I have noticed one dependency on ```ntpath``` in Julian de Wit's model which has to be replaced with ```os.path``` before it can run on Linux. This seems to be even more important with the Daniel Hammack algorithm (a.o. a possible openCV issue and Keras amendments that can cause problems in reproducing the results).

Is it important to have detailed versions of libraries for the documentation here or should be cross that bridge when we get there?

lamby:
Merged for now.

Issue #34, opened 2017-08-04
Add ability to set right/left lung for candidate nodule


lamby:
Add ability to set right/left lung for candidate nodule

## Overview

Annotation and segmentation view should support ability to save which right/left lung the nodule corresponds to; before we pass back any more detailed or complex information, we should simply provide a way to return this information so that adding further data in future is easier.

## Expected Behavior

It should be possible to select and subsequently save the per-nodule lung orientation using the dropdown:

![](https://i.imgur.com/lXmTg09.jpg)

Note that only the left/right lung selection is part of this issue.

## Technical details

The dropdown should not save until an "Accept" button has been pressed; see the design document for an example wireframe.

## Acceptance criteria

Per-nodule lung orientation should be returned and persisted to the backend service, once "Accept" is pressed.

## Blocked by
Issues: #9

PR #101, opened 2017-09-09
#34 Select Lung Orientation

WGierke:
This PR adds the ability to select whether a nodule belongs to the right or to the left lung.

## Description
I added an accordion view of all nodules with a drop-down menu to select in which part of the lung a nodule is located.

This addresses #34 .

## Motivation and Context
This should be a simple entry point for adding information to a nodule.

## How Has This Been Tested?
I wrote a [test](https://github.com/concept-to-clinic/concept-to-clinic/pull/101/files#diff-4a8b2a0cc0727d9b77d206172538ca85R36) that checks whether POSTing to the new back-end endpoint results in an appropriately updated nodule entity. Also, I tested it manually.

## Screenshots:
screenshot from 2017-09-09 23-05-11

## CLA


interface/backend/api/serializers.py
```diff
@@ -54,3 +54,11 @@ class Meta:
         read_only_fields = ('created',)
 
     centroid = ImageLocationSerializer()
+
+    def create(self, validated_data):
+        case_data = validated_data.pop('case')
```
lamby: (Similar to previous PR review; why pop, etc. etc.)

interface/backend/api/views.py
```diff
@@ -55,3 +59,13 @@ def candidate_mark(request, candidate_id):
 
 def candidate_dismiss(request, candidate_id):
     return JsonResponse({'response': "Candidate {} was dismissed".format(candidate_id)})
+
+
+@csrf_exempt
```
lamby: Why? (if justified, please add a comment the code so it's there forever!)

interface/backend/api/views.py
```diff
@@ -55,3 +59,13 @@ def candidate_mark(request, candidate_id):
...
+def nodule_update(request, nodule_id):
+    lung_orientation = json.loads(request.body)['lung_orientation']
```
lamby: Feel like we want some validation here or at least a ``try […] except`` \*g\*

interface/backend/cases/models.py
```diff
@@ -39,3 +51,5 @@ class Nodule(models.Model):
     candidate = models.OneToOneField(Candidate, on_delete=models.CASCADE, null=True)
 
     centroid = models.OneToOneField('images.ImageLocation', on_delete=models.CASCADE)
+
+    lung_orientation = models.CharField(max_length=1, choices=LungOrientations.choices(), default='0')
```
lamby: ``CharField`` smells wrong to me given that we can want to enforce numbers.

In addition the default should prbobably be ``LungOrientations.none`` rather than "0", if you see the semantic difference.

interface/backend/cases/tests.py
```diff
@@ -31,3 +32,26 @@ def test_create_nodule(self):
...
         self.assertIsInstance(nodule.centroid.y, int)
         self.assertIsInstance(nodule.centroid.z, int)
+
+    def test_update_nodule_lung_orientation(self):
+        nodule = NoduleFactory()
+        # Default lung orientation is none
+        self.assertEquals(nodule.lung_orientation, '0')
+
+        # Set lung orientation to left
+        self.client.post("/api/nodules/{}/update".format(nodule.id), json.dumps({'lung_orientation': 'left'}),
```
lamby: I think we should use ``reverse``` even here.

interface/backend/cases/utils.py
```diff
@@ -0,0 +1,14 @@
+import inspect
+from enum import Enum
+
+
+class ChoiceEnum(Enum):
```
lamby: There are a number of Django enum fields - did you look into them at all? They seem to do this stuff a little nicer, or at least with less bus-factor!

interface/frontend/annotate_and_segment.html
```diff
@@ -7,16 +7,106 @@
...
+            // error callback
+          });
+      },
+    }
+  })
+
+
+</script>
 {% endblock %}
+
```
lamby: Whitespace ;)

interface/backend/api/views.py
```diff
@@ -61,9 +60,11 @@ def candidate_dismiss(request, candidate_id):
     return JsonResponse({'response': "Candidate {} was dismissed".format(candidate_id)})
 
 
-@csrf_exempt
 def nodule_update(request, nodule_id):
-    lung_orientation = json.loads(request.body)['lung_orientation']
+    try:
+        lung_orientation = json.loads(request.body)['lung_orientation']
+    except:
+        return JsonResponse({'response': "An error occurred."})
```
lamby: This will return HTTP 200 now

interface/frontend/annotate_and_segment.html
```diff
@@ -7,16 +7,112 @@
...
+          function(response) {
+            this.nodules = response.body;
+          },
+          function() {
+            // error callback
+          });
+      },
+      update: function(nodule) {
+        this.$http.post(nodule.url + "update",
+          { csrfmiddlewaretoken: $("input[name=csrfmiddlewaretoken]").val(),
```
lamby: I wonder if we can do ths globally; kinda ugly to do it like this every time IMHO. jQuery has a global ajax handler- could you file a new issue?

WGierke: I filed issue #114 for that.

interface/backend/api/views.py
```diff
@@ -55,3 +58,15 @@ def candidate_mark(request, candidate_id):
 
 def candidate_dismiss(request, candidate_id):
     return JsonResponse({'response': "Candidate {} was dismissed".format(candidate_id)})
+
+
+def nodule_update(request, nodule_id):
+    try:
+        lung_orientation = json.loads(request.body)['lung_orientation']
+    except:
```
lamby: "Naked" ``try-except`` here? :)

interface/backend/cases/models.py
```diff
@@ -4,6 +4,18 @@
...
+
+class LungOrientations(ChoiceEnum):
+    none = 0
+    left = 1
+    right = 2
+
+    @classmethod
+    def reversed_choices_dict(self):
+        """Return the dictionary ORIENTATION->NUMBER"""
+        return dict([(y, x) for x, y in self.choices()])
```
lamby: Just ooi, have you seen the nifty:

```
{y: x for x, y in self.choices()}
```

syntax? :)

WGierke:
Thanks for the feedback @lamby! I already started to implement it. However, I'm a bit blocked by the fact that currently the Vue.js front-end neither has a router and other pages than the "Open Image" page, nor does jQuery seem to work (when trying to run `$('select[name=lung_orientation]').val()`, I'm getting the error `'$' is not defined`, when trying it with `window.$...` I'm getting a `window.$ is not a function`). @isms do you think you could have a look into that? I'm a complete novice when it comes to Vue :/ Otherwise I'd be supposed to wait for #110 to be completely implemented.

interface/frontend/annotate_and_segment.html
isms: Why would we use jQuery for this?

There's a simple way to set this up for Vue and Django:

```javascript
Vue.http.headers.common['X-CSRFToken']
```

Where `X-CSRFToken` is a header coming from Django.

antonow:
@WGierke and @isms, I would like to propose removing jQuery (and bootstrap.js) from the project altogether. Vue alone is able to handle all the front-end reactivity that bootstrap does in their JS-supported elements with less code. Also, jQuery's ajax isn't well suited to setting default data params -- see the convoluted solutions described [here](https://groups.google.com/forum/#!topic/jquery-dev/OBcEfgvTJ9I). Instead we could use [axios](https://github.com/mzabriskie/axios#request-config), which makes it easy to set base configuration options including options for `xsrfTokens` and `xsrfHeaders`. Or we could use Vue's built in http functionality as @isms suggested above.

Anyway, I am close to done with a PR that should address #110, and I've integrated/ported the frontend code from this PR for which I'll be sure to credit @WGierke. I hope to have that ready for review soon, but I wanted to give my opinion regarding using jQuery after reading through the most recent comments on this PR.

WGierke:
@antonow This sounds great since I'm a giant novice when it comes to Vue :) I removed my front-end code again and refactored the back-end code. Travis is still making some troubles although all tests pass locally ...

isms:
@antonow concur completely on removing jQuery.

Let's make the CSRF question even simpler: per the design document, security concerns are out of scope so XSS is not a concern at the moment.

I suggest we disable the CSRF middleware in Django. Thoughts?

lamby:
> I suggest we disable the CSRF middleware in Django. Thoughts?

Ew! Also very very hard to add it back later in my experience...

isms:
> Ew! Also very very hard to add it back later in my experience...

Fair, but this is out of scope. We specifically put auth out of bounds here. Is it worth a large effort to implement token based auth through DRF?

lamby:
> Is it worth a large effort to implement token based auth through DRF?

I'm.. not sure. Perhaps going the CSRF route is preferable overall, but I don't *quite* see what's so hard about keeping it? Just make sure we include a CSRF token on every page (or header) and let javascript always add it as a parameter? This is very very common in Django land.

WGierke:
Letting Django respond to every request with a CSRF token seems pretty easy by e.g. adding
`'config.middleware.ForceCsrfCookieMiddleware',` to `MIDDLEWARE` and adding 
```
from django.utils.deprecation import MiddlewareMixin
from django.middleware.csrf import get_token

class ForceCsrfCookieMiddleware(MiddlewareMixin):
    """
    Forces the CSRF cookie to be set for each request
    """
    def process_request(self, request):
        get_token(request)
```
in `interface/config/middleware.py`.
Regarding #125 , letting Vue (with axios) respond with the correct CSRF token seems to be as easy as 
```
import axios from 'axios';
axios.defaults.xsrfCookieName = 'csrftoken'
axios.defaults.xsrfHeaderName = 'X-CSRFToken'
```

However, is CSRF really something that should block this PR about handling the lung orientation in the back-end? Shouldn't the CSRF token management be added in another PR?

lamby:
@WGierke 
Thanks for looking into this. I knew the Django parts (and the associate jQuery global blah) but didn't know about the Vue part - thanks!

> is CSRF really something that should block this PR about handling the lung orientation in the back-end? Shouldn't the CSRF token management be added in another PR?

Haha, absolutely. Please open another issue at the very least and copy-paste what you wrote here into there? Of course, free free to come up with a PR too.. ;)

(AIUI it *was* blocking this...?)

WGierke:
@lamby I filed #134 . You're right, it blocked this PR when the front-end of the lung orientation selection was still part of this PR. Since that moved to #125 (and the issue has been filed), I'd say that this PR shouldn't be blocked anymore?

lamby:
Thanks! Rebased :)

Issue #111, opened 2017-09-13
Use Django Rest Framework responses consistently


isms:
## Expected Behavior
We shouldn't have `JsonResponse`s in the `backend` API. Let's follow the DRF idioms [1][2].

[1] http://www.django-rest-framework.org/api-guide/responses/
[2] http://www.django-rest-framework.org/tutorial/2-requests-and-responses/

## Current Behavior

```
$ rg Response
interface/backend/api/views.py
10:from rest_framework.response import Response
12:from django.http import JsonResponse
83:        return JsonResponse({'directories': tree})
87:    return JsonResponse({'response': "Candidate {} was marked".format(candidate_id)})
91:    return JsonResponse({'response': "Candidate {} was dismissed".format(candidate_id)})
```

PR #112, opened 2017-09-13
#111 Replace JsonResponses by Responses

WGierke:
I replaced JsonResponses by Responses which follows the DRF idioms.

This addresses #111 .

## How Has This Been Tested?
I added two tests so all functions, that previously used JsonResonses, are tested whether they still return valid JSON.

## CLA


lamby:
Thanks again @WGierke !

Issue #23, opened 2017-08-04
Document the Pierre Fillard (Therapixel) algorithm


reubano:
## Overview

Participants in the [Data Science Bowl](https://www.kaggle.com/c/data-science-bowl-2017) produced several algorithms that we would like to incorporate. To help facilitate this effort, we also want to add documentation so that contributors can make an educated decision when selecting an algorithm to incorporate.

## Expected Behavior

This documentation should enable people to:
 - view the library dependencies and license
 - understand its pros/cons
 - evaluate its performance/accuracy
 - identify which areas of the codebase to target for improvement

Design doc reference: [Detect and select](https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#detect-and-select)

## Algorithm info

key | value
----|------
team|Pierre Fillard (Therapixel)
rank|5
repo|https://github.com/pfillard/tpx-kaggle-dsb2017
trained models|https://github.com/pfillard/tpx-kaggle-dsb2017/tree/master/models
converted branch|
ML engine|
engine-version|
ML backend|Tensorflow
backend-version|1.1
training method|
architecture|
algorithm|
OS|Ubuntu
OS version|16.04
Python version|3.5
CUDA version|8
cuDNN version|5.1
notes|https://github.com/pfillard/tensorflow/tree/r1.0_relu1

## Technical details

 - This feature should be implemented as a new markdown file in the [`docs folder`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/docs)

## Acceptance criteria

 - [ ] effective documentation for the above

*NOTE: All PRs must follow the standard PR checklist.*

PR #115, opened 2017-09-14
#23 Document the Therapixel algorithm

WGierke:
I added an initial documentation of the Therapixel algorithm. The algorithm scored 5th place during the DSB 2017.

Unfortunately, there is no technical report or blog post about the approach so far.
Thus, I had to derive the algorithm design from the source code and black-box models only, which resulted in a shorter documentation than for other algorithms.

This pull request addresses #23 

## Motivation and Context

## CLA

### DIFF
--- a/docs/algorithm_therapixel.md
+++ b/docs/algorithm_therapixel.md
@@ -1,0 +1,151 @@
+# Therapixel
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


lamby:
Thanks!

Issue #116, opened 2017-09-15
small DICOM test images no longer managed in git


rahit:
As of #95, DICOM images in `tests/assets/test_image_data/small` were moved from `git-lfs` directly to `git`. They now appear to be included in `git-lfs`.

## Expected Behavior

DICOM images in `tests/assets/test_image_data/small` should be directly managed `git`

## Current Behavior

DICOM images in `tests/assets/test_image_data/small` are linked via `git-lfs`

## Possible Solution

This appears to be a bug introduced by commit [#740af48](https://github.com/concept-to-clinic/concept-to-clinic/commit/740af48fc8fc248fa47242f3e0ea39db081cb71b) from #76.  I suspect the culprit is [somewhere here in `.gitattributes`](https://github.com/concept-to-clinic/concept-to-clinic/pull/76/files#diff-fc723d30b02a4cca7a534518111c1a66).

## Steps to Reproduce

```bash
git checkout 9b1c31
cd tests/assets/test_image_data/small/LIDC-IDRI-0001/
file  '1.3.6.1.4.1.14519.5.2.1.6279.6001.298806137288633453246975630178/1.3.6.1.4.1.14519.5.2.1.6279.6001.179049373636438705059720603192/-100.000000.dcm'
```

`output`
`1.3.6.1.4.1.14519.5.2.1.6279.../-100.000000.dcm: DICOM medical imaging data`

```bash
git checkout 740af4
file  '1.3.6.1.4.1.14519.5.2.1.6279.6001.298806137288633453246975630178/1.3.6.1.4.1.14519.5.2.1.6279.6001.179049373636438705059720603192/-100.000000.dcm'
```
`output`
`1.3.6.1.4.1.14519.5.2.1.6279.../-100.000000.dcm: ASCII text`

I encountered this while trying to load metadata information as of #33 

## Possible Implementation

@Serhiy-Shekhovtsov @vessemer any ideas?

## Checklist before submitting
- [x] I have confirmed this using the officially supported Docker Compose setup using the `local.py` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [x] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [x] I provided a **minimal code snippet** or list of steps that reproduces the bug.
- [x] I provided **screenshots** where appropriate
- [x] I filled out all the relevant sections of this template

WGierke:
[...]

Issue #33, opened 2017-08-04
Display metadata about an image prior to selection


lamby:
## Overview

When browsing for an test case to analyse, it would be very helpful for the end-user to be able to preview a particular case before proceeding to the next step. This would also provide means to ensure and double-check they are selecting the correct patient.

We should therefore display metadata about an image prior to selection.

## Expected Behavior

The image metadata should be displayed in a panel on the right-hand side of the interface. For example:

![](https://i.imgur.com/RPBh9wt.jpg)

## Technical details

Upon clicking on an image file, the data will have to be loaded. A preview of the image will need to be displayed too, likely in a seperate HTTP call, but it may be possible and cleaner to provide a preview within the same payload by base64-encoding the image data.

## Acceptance criteria

A pane on the right-hand side of the open image will reveal all the relevant metadata for an image.

## Blocked by
Issues: #9

WGierke:
Do you know what the "mA" (probably milliampere?) information means and how to access it using `pydicom`?

WGierke:
@lamby friendly ping :)

lamby:
Fixed in #117!

PR #117, opened 2017-09-16
33 DICOM metadata API

rahit:
API implementation for dicom metadata and preview available at:
`/api/images/metadata?dicom_location='IMAGE_PATH'`.

## Description
Before selecting particular DICOM image for analysis we need to display its metadata with preview. This API provides those data.

#33 

## How Has This Been Tested?
Test case to check HTTP 200 OK status for valid dicom image has been added. However, because of the issue #116, test cases are failing for the test images. I am adding screenshot which I verified using a dicom image from original dataset.

image metadata api django rest framework

## CLA


interface/backend/api/views.py
```diff
@@ -42,6 +46,78 @@ class ImageSeriesViewSet(viewsets.ModelViewSet):
     serializer_class = serializers.ImageSeriesSerializer
 
 
+class ImageMetadataApiView(APIView):   
+
+    def _sanitise_unicode(self, s):
+        return s.replace(u"\u0000", "").strip()
+
+    def _convert_value(self, v):
+        t = type(v)
```
lamby: The typical way to do this in Python is using ``isinstance``.

interface/backend/api/views.py
```diff
@@ -42,6 +46,78 @@ class ImageSeriesViewSet(viewsets.ModelViewSet):
...
+            cv = float(v)
+        elif t == dicom.valuerep.IS:
+            cv = int(v)
+        elif t == dicom.valuerep.PersonName3:
+            cv = str(v)
+        else:
+            cv = repr(v)
+        return cv
+
+    def dicom_dataset_to_dict(self, ds):
```
lamby: Hm, I think Django has some kind of idea about a serializer? Perhaps we should extend support via that instead of these custom routines? Alternatively, we could extend Django's JSON serializer? Anyway I think we should have this in ``utils.py`` or similar anyway :)

interface/backend/api/views.py
```diff
@@ -42,6 +46,78 @@ class ImageSeriesViewSet(viewsets.ModelViewSet):
...
+        '''
+        Put dicom metadata into a separate dictionary
+        '''
+        dicom_dict = {}
+        repr(ds)
```
lamby: ?

interface/backend/api/tests.py
```diff
@@ -41,6 +41,15 @@ def test_images_available_view(self):
...
         self.assertEqual(response.status_code, status.HTTP_200_OK)
+        
+    def test_images_metadata_view(self):
+        url = reverse('images-metadata')
+        response = self.client.get(url, {
+            'dicom_location': '"/images/LIDC-IDRI-0002/' \
+            '1.3.6.1.4.1.14519.5.2.1.6279.6001.490157381160200744295382098329/' \
+            '1.3.6.1.4.1.14519.5.2.1.6279.6001.619372068417051974713149104919/-80.750000.dcm"'
+        })
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
WGierke: Would it make sense to test for some specific metadata attributes that should be correct and are known for the image? Like SeriesInstanceUID?

interface/backend/api/views.py
isms: Django Rest Framework Serializer: http://www.django-rest-framework.org/api-guide/serializers/

interface/backend/api/views.py
```diff
@@ -42,6 +43,27 @@ class ImageSeriesViewSet(viewsets.ModelViewSet):
...
+        parameters:
+            - name: dicom_location
+            description: 'QUOTED' full location of the image
+            required: true
+            type: string
+        '''
+        path = request.GET.get('dicom_location')
+        if path is None:
+            raise Exception('dicom_location not provided')
+        path = path[1:-1]   # un-quoting the string
```
lamby: Hang on, why would we be quoted?

lamby: (Please clarify in docstring for posterity, not here....)

rahit: @lamby  unquoted path seems to create problem with existing url patterns.

lamby: "seems to"?

And again, please clarify the reason in the docstring itself, otherwise these important comments just get lost :)

interface/backend/api/views.py
```diff
@@ -42,6 +43,27 @@ class ImageSeriesViewSet(viewsets.ModelViewSet):
...
+        ---
+        parameters:
+            - name: dicom_location
+            description: 'QUOTED' full location of the image
+            required: true
+            type: string
+        '''
+        path = request.GET.get('dicom_location')
+        if path is None:
+            raise Exception('dicom_location not provided')
```
lamby: I don't recommend raising ``Exception`` - there is almost always a better one. Indeed, why not just do a raw ``request.GET`` and let it ``KeyError`` and — potentially — reraise it as a ``ValueError``. I mean, that's going to be far more useful to the callsite.

lamby:
@rahit As I understand it, this is waiting on you? :)

rahit:
@lamby sorry for the late.

PR #118, opened 2017-09-17
#1 Identify Nodules: Julian de Wit Algorithm

WGierke:
I adapted the source code of [Julian de Wit](https://github.com/juliandewit/kaggle_ndsb2017) whose solution scored 2nd pace during the DSB 2017. I decided to use it as it was already written for Python 3, has a manageable code base and immediately ran using my CPU only. The algorithm is documented [here](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/docs/algorithm_de_wit.md). It uses a Conv3D network to detect nodules. The network returns a set of nodules characterized by their position and the probability, that the marked tissue really is a nodule. This output is wrapped into the placeholder function that's already implemented in [prediction/src/algorithms/identify/trained_model/predict](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/prediction/src/algorithms/identify/trained_model.py#L11-L38).

## Description
I took most of the code from steps 2 (preprocessing) and 4 (predicting) of De Wits repository as well as the models. The models, that were trained on different data sets with different data transformations such as magnification, are saved using Git LFS.
So far, I kept most of the commented code so it's easier to compare this ported version with his code, especially since his implementation is not very well documented what makes it difficult to understand why certain code regions are important.

This addresses #1 

## Motivation and Context
We need a possibility to automatically find nodules so we can calculate further whether they are malicious or not. 
I believe that it might be even better to have multiple algorithms that detect nodules and their malice since this is supposed to be the core feature of this whole project.

## How Has This Been Tested?
It takes approximately 8 Minutes to detect nodules of a full DICOM scan using CPU only with my i7. ~~Thus, I didn't want to write a test (as it probably would take days for that to finish on Travis) so I can only provide a screenshot of the output of the API endpoint:~~
Edit: I added a decorator that skips test functions if [`RUN_SLOW_TESTS`](https://github.com/concept-to-clinic/concept-to-clinic/pull/118/files#diff-5403e981dddb4b13174576f39e11b935R64) is set to 0. By setting the environment variable to 1 you can also run the [test](https://github.com/concept-to-clinic/concept-to-clinic/pull/118/files#diff-39dfe0cb4eefe4ddee11016381c0a850R89) locally.
screenshot from 2017-09-18 00-41-46

I'd highly appreciate any feedback how to further simplify and beautify the code.

## CLA

### DIFF
--- a/local.yml
+++ b/local.yml
@@ -64,6 +64,9 @@ services:
     volumes:
       - ./prediction/:/app
       - ./tests/assets/test_image_data/small:/images
+      - ./tests/assets/test_image_data/full:/images_full
       - ./prediction/src/algorithms/classify/assets/:/classify_models
+      - ./prediction/src/algorithms/identify/assets/:/identify_models
+      - ./data/:/app/data
     ports:
       - "8001:8001"
--- a/prediction/src/algorithms/identify/helpers.py
+++ b/prediction/src/algorithms/identify/helpers.py
@@ -1,0 +1,175 @@
+import glob
+import logging
+import os
+from collections import defaultdict
+
+import cv2
+import dicom
+import numpy
+import scipy
+from skimage.filters import roberts
+from skimage.measure import label, regionprops
+from skimage.morphology import disk, binary_erosion, binary_closing
+from skimage.segmentation import clear_border
+
+PATIENT_LIST = None
+PRINT_TAB_MAP = defaultdict(lambda: [])
+
+
+def load_patient_images(patient_id, base_dir=None, wildcard="*.*", exclude_wildcards=None):
+    exclude_wildcards = exclude_wildcards or []
+    src_dir = base_dir + patient_id + "/"
+    src_img_paths = glob.glob(src_dir + wildcard)
+    for exclude_wildcard in exclude_wildcards:
+        exclude_img_paths = glob.glob(src_dir + exclude_wildcard)
+        src_img_paths = [im for im in src_img_paths if im not in exclude_img_paths]
+    src_img_paths.sort()
+    images = [cv2.imread(img_path, cv2.IMREAD_GRAYSCALE) for img_path in src_img_paths]
+    images = [im.reshape((1,) + im.shape) for im in images]
+    res = numpy.vstack(images)
+    return res
+
+
+def normalize_hu(image):
+    MIN_BOUND = -1000.0
+    MAX_BOUND = 400.0
+    image = (image - MIN_BOUND) / (MAX_BOUND - MIN_BOUND)
+    image[image > 1] = 1.
+    image[image < 0] = 0.
+    return image
+
+
+def rescale_patient_images(images_zyx, org_spacing_xyz, target_voxel_mm, is_mask_image=False):
+    resize_x = 1.0
+    resize_y = org_spacing_xyz[2] / target_voxel_mm
+    interpolation = cv2.INTER_NEAREST if is_mask_image else cv2.INTER_LINEAR
+    res = cv2.resize(images_zyx, dsize=None, fx=resize_x, fy=resize_y,
+                     interpolation=interpolation)  # opencv assumes y, x, channels umpy array, so y = z pfff
+
+    res = res.swapaxes(0, 2)
+    res = res.swapaxes(0, 1)
+
+    resize_x = org_spacing_xyz[0] / target_voxel_mm
+    resize_y = org_spacing_xyz[1] / target_voxel_mm
+
+    # cv2 can handle max 512 channels..
+    if res.shape[2] > 512:
+        res = res.swapaxes(0, 2)
+        res1 = res[:256]
+        res2 = res[256:]
+        res1 = res1.swapaxes(0, 2)
+        res2 = res2.swapaxes(0, 2)
+        res1 = cv2.resize(res1, dsize=None, fx=resize_x, fy=resize_y, interpolation=interpolation)
+        res2 = cv2.resize(res2, dsize=None, fx=resize_x, fy=resize_y, interpolation=interpolation)
+        res1 = res1.swapaxes(0, 2)
+        res2 = res2.swapaxes(0, 2)
+        res = numpy.vstack([res1, res2])
+        res = res.swapaxes(0, 2)
+    else:
+        res = cv2.resize(res, dsize=None, fx=resize_x, fy=resize_y, interpolation=interpolation)
+
+    res = res.swapaxes(0, 2)
+    res = res.swapaxes(2, 1)
+
+    return res
+
+
+def rescale_patient_images2(images_zyx, target_shape):
+    interpolation = cv2.INTER_LINEAR
+    res = cv2.resize(images_zyx, dsize=(target_shape[1], target_shape[0]),
+                     interpolation=interpolation)  # opencv assumes y, x, channels umpy array, so y = z pfff
+
+    res = res.swapaxes(0, 2)
+    res = res.swapaxes(0, 1)
+
+    # cv2 can handle max 512 channels..
+    if res.shape[2] > 512:
+        res = res.swapaxes(0, 2)
+        res1 = res[:256]
+        res2 = res[256:]
+        res1 = res1.swapaxes(0, 2)
+        res2 = res2.swapaxes(0, 2)
+        res1 = cv2.resize(res1, dsize=(target_shape[2], target_shape[1]), interpolation=interpolation)
+        res2 = cv2.resize(res2, dsize=(target_shape[2], target_shape[1]), interpolation=interpolation)
+        res1 = res1.swapaxes(0, 2)
+        res2 = res2.swapaxes(0, 2)
+        res = numpy.vstack([res1, res2])
+        res = res.swapaxes(0, 2)
+    else:
+        res = cv2.resize(res, dsize=(target_shape[2], target_shape[1]), interpolation=interpolation)
+
+    res = res.swapaxes(0, 2)
+    res = res.swapaxes(2, 1)
+    return res
+
+
+def get_segmented_lungs(im):
+    # Step 1: Convert into a binary image.
+    binary = im < -400
+    # Step 2: Remove the blobs connected to the border of the image.
+    cleared = clear_border(binary)
+    # Step 3: Label the image.
+    label_image = label(cleared)
+    # Step 4: Keep the labels with 2 largest areas.
+    areas = [r.area for r in regionprops(label_image)]
+    areas.sort()
+    if len(areas) > 2:
+        for region in regionprops(label_image):
+            if region.area < areas[-2]:
+                for coordinates in region.coords:
+                    label_image[coordinates[0], coordinates[1]] = 0
+    binary = label_image > 0
+    # Step 5: Erosion operation with a disk of radius 2.
+    # This operation is seperate the lung nodules attached to the blood vessels.
+    selem = disk(2)
+    binary = binary_erosion(binary, selem)
+    # Step 6: Closure operation with a disk of radius 10.
+    # This operation is to keep nodules attached to the lung wall.
+    selem = disk(10)  # CHANGE BACK TO 10
+    binary = binary_closing(binary, selem)
+    # Step 7: Fill in the small holes inside the binary mask of lungs.
+    edges = roberts(binary)
+    binary = scipy.ndimage.binary_fill_holes(edges)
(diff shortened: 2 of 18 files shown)


lamby:
@WGierke 

> It takes approximately 8 Minutes to detect nodules of a full DICOM scan using CPU only with my i7. Thus, I didn't want to write a test (as it probably would take days for that to finish on Travis)

Good thoughts.

However, what I suggest we do is to make a decorator for tests thats skip them *except* for the case where we export an environment variable such as ``RUN_SLOW_TESTS=1``. That way we can have almost the best of both worlds. Don't forget that tests also function as kind of documentation/examples on how to use an API, so we would gain that useful bit as well. :)

WGierke:
Nice idea @lamby . I implemented that in c3c626f

prediction/src/tests/test_endpoints.py
```diff
@@ -4,14 +4,35 @@
...
+    """
+    Skip the wrapped test function unless the environment variable RUN_SLOW_TESTS is set to 1.
+    """
+
+    def wrapper(func):
+        @wraps(func)
+        def wrapped(*args, **kwargs):
+            import os
+            if os.environ.get('RUN_SLOW_TESTS') != '1':
+                return
```
lamby: This should throw some kind of ``skip`` exception, not silently return.

local.yml
```diff
@@ -61,9 +61,13 @@ services:
     environment:
       - FLASK_DEBUG=1
       - FLASK_APP=src/factory.py
+      - RUN_SLOW_TESTS=0
```
lamby: No need for this; surely it's implicit?

Adding it here actually implies it's *required* otherwise we run the slow tests.

prediction/src/tests/test_endpoints.py
```diff
@@ -4,14 +4,35 @@
...
+            return func(*args, **kwargs)
+
+        return wrapped
+
+    if func:
+        return wrapper(func)
+    else:
```
lamby: No need for this ``else`` and indented, surely?

prediction/src/tests/test_endpoints.py
WGierke: Maybe I got you wrong but when I simply put a `raise Exception(...)` there, the tests fails. If we try to print or log something, it's captured by pytest und only shown if the respective test fails. So what about normal printing but running pytest with a -s flag, which does not capture stdout but would also show warnings and exception messages of other tests that occur when running them?
It would currently look like this:
```
(venv3.5) ➜  concept-to-clinic git:(1_identify_nodules) ✗ sudo bash tests/test_docker.sh
+ docker-compose -f local.yml run prediction pytest -s
======================================== test session starts ========================================
platform linux -- Python 3.6.2, pytest-3.1.3, py-1.4.34, pluggy-0.4.0
rootdir: /app, inifile:
collecting 3 itemsUsing TensorFlow backend.
collected 22 items 

src/tests/test_calculate_volume.py ...
src/tests/test_classify_trained_model_predict.py .2017-09-19 07:33:33.828938: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.1 instructions, but these are available on your machine and could speed up CPU computations.
2017-09-19 07:33:33.828960: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.2 instructions, but these are available on your machine and could speed up CPU computations.
2017-09-19 07:33:33.828965: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use AVX instructions, but these are available on your machine and could speed up CPU computations.
2017-09-19 07:33:33.828969: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use AVX2 instructions, but these are available on your machine and could speed up CPU computations.
2017-09-19 07:33:33.828973: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use FMA instructions, but these are available on your machine and could speed up CPU computations.
.
src/tests/test_crop_dicom.py .
src/tests/test_endpoints.py ..test_identify has been skipped
......
src/tests/test_load_dicom.py Exception reading *.dcm-files:  File is missing 'DICM' marker. Use force=True to force reading
.Exception extracting voxel data:  Dataset does not have attribute 'ImageOrientationPatient'.
...
src/tests/test_preprocess_dicom.py ....

==================================== 22 passed in 27.18 seconds =====================================
+ docker-compose -f local.yml run interface python manage.py test
WARNING: Found orphan containers (concepttoclinic_compile_docs_1) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.
Starting concepttoclinic_postgres_1 ... 
Starting concepttoclinic_prediction_1 ... done
Postgres is up - continuing...
/usr/local/lib/python3.6/site-packages/environ/environ.py:618: UserWarning: Error reading /app/.env - if you're not configuring your environment separately, check this.
  "environment separately, check this." % env_file)
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
..........
----------------------------------------------------------------------
Ran 10 tests in 0.073s

OK
Destroying test database for alias 'default'...
+ docker-compose -f local.yml run interface python manage.py makemigrations --dry-run --check
WARNING: Found orphan containers (concepttoclinic_compile_docs_1) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.
Starting concepttoclinic_postgres_1 ... 
Starting concepttoclinic_postgres_1 ... done
Postgres is up - continuing...
/usr/local/lib/python3.6/site-packages/environ/environ.py:618: UserWarning: Error reading /app/.env - if you're not configuring your environment separately, check this.
  "environment separately, check this." % env_file)
No changes detected

```

local.yml
WGierke: I thought it would make sense to add it here so people directly know where to change that "switch".

prediction/src/tests/test_endpoints.py
reubano: So you can you pytest here to help you

```python
import pytest

def skip_slow_test():
    return not os.environ.get('RUN_SLOW_TESTS')


@pytest.mark.skipif(skip_slow_test, reason='some reason')
def test_slow_test():
    pass
```

prediction/src/algorithms/identify/helpers.py
```diff
@@ -0,0 +1,174 @@
...
+        try:
+            slices.append(dicom.read_file(src_dir + '/' + s))
+        except Exception as e:
+            logging.error("{} no valid DICOM: {}".format(s, e))
+    slices.sort(key=lambda x: int(x.InstanceNumber))
+    try:
+        slice_thickness = numpy.abs(slices[0].ImagePositionPatient[2] - slices[1].ImagePositionPatient[2])
+    except Exception as e:
+        slice_thickness = numpy.abs(slices[0].SliceLocation - slices[1].SliceLocation)
+
```
reubano: I think it'd be better to catch the specific exceptions in these cases. Using bare exceptions is usually code smell unless you plan to log and then reraise it.

prediction/src/tests/test_endpoints.py
WGierke: Thanks @reubano that's way cleaner! When handing over `-rs` to pytest, the skip reason is shown without the output of other tests being displayed. Nice!

prediction/src/algorithms/identify/prediction.py
```diff
@@ -0,0 +1,232 @@
...
+
+def prepare_image_for_net3D(img):
+    img = img.astype(numpy.float32)
+    img -= MEAN_PIXEL_VALUE
+    img /= 255.
+    img = img.reshape(1, img.shape[0], img.shape[1], img.shape[2], 1)
+    return img
+
+
+def filter_patient_nodules_predictions(df_nodule_predictions: pandas.DataFrame, patient_id, view_size):
```
reubano: is that first `:` a typo? :)

reubano: hmm, possible this is a typehint? I must admit, I haven't used those yet... heh.

prediction/src/algorithms/identify/trained_model.py
```diff
@@ -34,8 +48,111 @@ def predict(dicom_path):
...
     """
-    load_dicom(dicom_path)
-    return [{'x': 0,
-             'y': 0,
-             'z': 0,
-             'p_nodule': 0.5}]
+    if dicom_path[-1] != '/':
+        dicom_path += '/'
+    dicom_files = glob.glob(dicom_path + "*.dcm")
+    if len(dicom_files) == 0:
```
reubano: if `dicom_files` is a plain iterable, `if not dicom_files:` works (and is more pythonic as well :))...

prediction/src/algorithms/identify/prediction.py
WGierke: Yes, it's a typehint. The author also used it e.g. in `def get_net(input_shape=(CUBE_SIZE, CUBE_SIZE, CUBE_SIZE, 1), load_weight_path=None) -> Model:`.
Since that has never been used in our project, I'll remove it again for consistency reasons?

reubano: Actually, that `->` is what triggered me recognizing it. I don't think you need to remove them per se, but IMHO type hints would be better suited "in addition" to doc blocks as opposed to "in stead of".

prediction/src/tests/test_endpoints.py
```diff
@@ -4,14 +4,21 @@
...
-from src.factory import create_app
 from src.algorithms import classify, identify, segment
+from src.factory import create_app
+
+
+def skip_slow_test():
+    """
+    Skip the wrapped test function unless the environment variable RUN_SLOW_TESTS is set.
+    """
+    return os.environ.get('RUN_SLOW_TESTS') != '1'
```
reubano: In cases like this, I think `not os.environ.get('RUN_SLOW_TESTS')` is more flexible as it doesn't require the value to be strickly `'1'`. Also, `'0'` is a boolean `True`, so it should probably be set to an int ;).

reubano: Just to clarify... any the following could be valid `RUN_SLOW_TESTS` values

`True`
`1`
`'any non empty string'`

prediction/src/tests/test_endpoints.py
```diff
@@ -18,7 +18,7 @@ def skip_slow_test():
     """
     Skip the wrapped test function unless the environment variable RUN_SLOW_TESTS is set.
     """
-    return os.environ.get('RUN_SLOW_TESTS') != '1'
+    return not os.environ.get('RUN_SLOW_TESTS')
```
reubano: Ok... you're gonna hate me for this... I think I spoke too soon. I'm use to python configs which get parsed to python objects. OS env values are returned as strings so you were right!! Apologies. :) The remaining question though becomes how to handle non `'1'` values. Since `local.yml` is a yaml file, I think some people would expect to be able to enter `True` as well. Borrowing from what I [read here](https://stackoverflow.com/a/43087058/408556) something like this may work.

```python 
value = os.environ.get('RUN_SLOW_TESTS')
return value.lower() not in {'1', 'true'}
```

Thoughts??

WGierke: When I see that there is a `RUN_SLOW_TESTS=0` in the YAML file and I'd like to run the slow tests, I'd change it to `RUN_SLOW_TESTS=1`. On the other hand, adding one line of code shouldn't be labeled "over-engineered" IMO :)

tests/test_docker.sh
```diff
@@ -7,7 +7,7 @@
 set -ex
 
 # run the model service's tests
-docker-compose -f local.yml run prediction pytest
+docker-compose -f local.yml run prediction pytest -rs
```
lamby: What does ``-rs`` do? :)

WGierke:
It displays a summary of skipped tests :) otherwise, every attempt to print something by a test is captured and only showed if the test fails.

tests/test_docker.sh
WGierke: It displays a summary of skipped tests :) otherwise, every attempt to print something by a test is captured and only showed if the test fails. [See the doc reference](https://docs.pytest.org/en/latest/example/simple.html)

lamby: Neat!

lamby:
@reubano LGTM. You?

reubano:
:shipit:

Issue #98, opened 2017-09-07
Add function that loads MetaImage files


vessemer:
In the field of low-dose CT, not the only DICOM data format is wildly used, but MetaImage (`.mhd` / `.raw`) is also taken place. The MetaImage is the text-based tagged file format for medical images. As well as the DICOM format, the MetaImage has descriptors along with the raw data.

## Expected Behavior
The `.mhd` / `.raw` files may be read via `SimpleITK`. So `load_itk` should be provided which takes the `itk_path` as an input and return the `np.ndarray` (right now the `load_dicom` returns the np.ndarray which may be different in the future since MetaImage and DICOM have different metadata and it will be the good decision to build a class which will standardise the metadata by parameters). Furthermore, the orchestrate function `load_CT` aimed at handling different medical image formats, so the `load_CT` will take `CT_path` as an input and will decide whether to return the output of `load_dicom` or `load_itk`. 

## Current Behavior
For now, exists only the possibility to read DICOM file format through `load_dicom`. 

## Technical details
This feature should be implemented in the `prediction/src/preprocess` folder

## Acceptance criteria
- [ ] method to load MetaImage (`.mhd` / `.raw`) images into memory
- [ ] method to select apropriate loader function
- [ ] tests for the methods

tdraebing:
I would start working on that. Are there any image files in those formats in the larger data set that will be used for model training/testing later on that could be used for tests? Else I would try to find some.

vessemer:
As far as I saw no changes for more than the week, I've created a pull request.

PR #119, opened 2017-09-18
Add function that loads MetaImage files

vessemer:
The `load_metaimage` function that load MetaImage files have been proposed along with `load_ct` ensembling method which aimed to organize loading process. Refactoring has been performed, taking into account various format of meta-information. In order to handle it standardisation via `class MetaData` has been employed, for now, it contains only one parameter (`spacing`) which is used already. Also decoupling of reading and preprocessing was included.

## Motivation and Context
This addresses [#98](https://github.com/concept-to-clinic/concept-to-clinic/issues/98). The MetaImage file format is actively used by plenty of medical studies. The LUNA16 challenge is among them. Since LUNA16 was involved in kaggle DSB 2017 competition, therefore, it's into of our area of interest. 

## How Has This Been Tested?
The unit tests were created for each of introduced functions.

## CLA


tdraebing:
Hi @vessemer,

thanks for the pull request. I am currently pretty swamped with the last months in my PhD, so I didn't manage to finish my pull request in the last days.
Based on my work on the issue, I have some further suggestions:
- The `load_dicom`-function uses pydicom and dicom-numpy to load the DICOM-images. SimpleITK also has a method implemented to do so (which might even be more tolerant to compressed images). Fully switching to SimpleITK would reduce the number of dependencies and make the code easier to maintain and follow. There would be also some changes in `crop_dicom` and `preprocess_dicom` needed. This might also be handled in a new issue.
- In your `MetaData`-class you calculate the slice thickness. There actually is a value for this in the metadata of the DICOM-images: '0018|0050'. (or for a pydicom-object `SliceThickness`.
- The `load_meta`-function seems to be quite redundant to `load_ct`. Maybe by adding a flag (like `voxel=True` if the voxel data should be returned as well) would be better here.
- This is just my personal preference, but maybe it would make sense to put the logic in `MetaData.__init__()` into a separate method(s). That would make it easier to extend the object later on. For example if patient data becomes relevant in the interface, which would also be stored in the metadata.
- From your tests I read that you added data from the LUNA16-competition. Did you check, whether using them in other projects is fine with their license?

dchansen:
I would strongly support moving to SimpleITK for everything, as it correctly deals with things like calibration curves, orientations, etc and gives us an easy way to pass along images and their metadata in one package.

vessemer:
Hi @tdraebing, thanks for your detailed review and suggestions. I've refactored the code, all rebased and stashed.

> From your tests I read that you added data from the LUNA16-competition. Did you check, whether using them in other projects is fine with their license?

As it was claimed on the official page of [LUNA16/data](https://luna16.grand-challenge.org/data/) the dataset was made from  LIDC/IDRI database via filtering out CT scans which slice thickness greater than 2.5 mm and may be other processing. The LIDC/IDRI database, in turn, consist of DICOM files and have [Creative Commons Attribution 3.0 Unported License](https://creativecommons.org/licenses/by/3.0/). Thus it's fine to use the LUNA16 data in a project. 

> Fully switching to SimpleITK would reduce the number of dependencies and make the code easier to maintain and follow. There would be also some changes in crop_dicom and preprocess_dicom needed. This might also be handled in a new issue.

I guess it will be a good idea to open a new issue dedicated to that task since it will lead to a quite cumbersome refactoring. @tdraebing, would you like to open it?

tdraebing:
Awesome. Thanks for checking up on the license! I opened a new issue to deal with loading the DICOM-images with SimpleITK.

lamby:
Great stuff guys! Thanks :)

Issue #1, opened 2017-07-29
Feature: Implement identification algorithm


pjbull:
## Overview
We need to adapt the Data Science Bowl algorithms to produce possible centroid locations for nodules within an image rather than just `P(cancer)` for the whole image.

## Expected Behavior
Currently, there is a just a placeholder in the algorithm that identifies nodules in scans. Nodules are areas of interest that might be cancerous (or might not be, the goal here is just the potentially concerning areas). This must actually yield centroid locations of potential nodules (X voxels from left, Y voxels from top, Z slice number).

First we need to train a model to perform this task. Then, we need to serialize the model so that it can be loaded from disk and used to make predictions. This trained model should be added to the `prediction/src/algorithms/identify/assets/` folder using `git-lfs`. Finally, we need to write the code in the `predict` method that will load the model from assets, take in a DICOM image, and yield nodule locations in the specified format.

Design doc reference:
[Jobs to be done > Detect and select > Prediction service](https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#prediction-service)

## Technical details
 - This feature should be implemented in the [`prediction/src/algorithms/identify/trained_model/predict`](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/prediction/src/algorithms/identify/trained_model.py#L11-L38) method
 - Code to train the model should live in the [`prediction/src/algorithms/identify/src/`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/prediction/src/algorithms/identify/src) folder
 - A fully serialized version of the model that can be loaded should live in the [`prediction/src/algorithms/identify/assets/`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/prediction/src/algorithms/identify/assets) folder

## Out of scope

This feature is a first-pass at getting a model that completes the task with the defined input and output. We are not yet judging the model based on its accuracy or computational performance.

## Acceptance criteria
 - [ ] trained model for identification
 - [ ] documentation for the trained model (e.g., cross validation performance, data used) and how to re-train it

*NOTE: All PRs must follow the standard PR checklist.*

QuantumDamage:
Hi @pjbull ,
is there any data set with already marked nodules which can be used to training? As far as I checked, Data Science Bowl has only data related to class, not actual nodules.

pjbull:
[...]

Issue #4, opened 2017-08-01
Feature: Adapt the grt123 model


reubano:
## Overview

Currently, there is a just a placeholder in the algorithm that classifies nodules in scans. Nodules are areas of interest that might be cancerous. We need to adapt the Data Science Bowl (DSB) algorithms to predict `P(cancer)` given an iterator of nodule centroids for an image. 

The top DSB algorithm (grt123) was written to run on a GPU for Python2. It would be nice to integrate this algorithm into the current structure and update it to run on Python3 (potentially on a CPU as well).

## Expected Behavior

Given the grt123 model trained to perform this task, a DICOM image, and an iterator of nodule centroids, yield the `P(cancer)` for each nodule.

Design doc reference: [Detect and select](https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#detect-and-select)

## Technical details

- The majority of the Python3 and CPU conversion has been completed and is available in the [conversion branch](https://github.com/concept-to-clinic/DSB2017/tree/conversion).

- The forked model is [available here](https://github.com/concept-to-clinic/DSB2017) (reads source DICOM images from S3).

- One area that definitely needs review is the Py2/Py3 floor/true division conversions. Some calculation explicitly converted numbers to floats, and in those cases it was apparent that true division was desired. However, the remaining floor division calculations should be checked to ensure that true is not the appropriate operation.

- If you get a `UnicodeDecodeError` error while trying to load the serialized Torch model, use the `torch_loader` function in the `utils` module instead.    

- When running on the CPU, it isn't necessary to perform that much work. Just enough to obtain a plausible result in a reasonable amount of time.

- This feature should be implemented in the `prediction/classify/trained_model/predict` method.

## Acceptance criteria

 - [ ] the integrated model produces results similar to [the forked model](https://github.com/concept-to-clinic/DSB2017)

*NOTE: All PRs must follow the standard PR checklist.*

Serhiy-Shekhovtsov:
I'd like to work on this one.

P.S. btw, is it a good idea to tell when you are working on something?

Serhiy-Shekhovtsov:
Hi @reubano! I've fixed few issues on pre-processing part and managed to have it running successfully and giving the same result as original code. These changes have been pushed now. 
Now I am looking into detection part, but encountered this issue:

> Torch: not enough memory: you tried to allocate 21GB

Does it really take this much memory or is it an issue with environment \ code?

reubano:
[...]

Issue #2, opened 2017-08-01
Feature: Implement classification algorithm


reubano:
## Overview

Currently, there is a just a placeholder in the algorithm that classifies nodules in scans. Nodules are areas of interest that might be cancerous. We need to adapt the Data Science Bowl algorithms to predict `P(cancer)` for a given set of centroids for nodules.

## Expected Behavior

Given a model trained to perform this task, a DICOM image, and the nodule centroids, return the `P(cancer)` for each nodule.

Design doc reference:
[Jobs to be done > Annotate > Prediction service](https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#prediction-service)

## Technical details

- This feature should be implemented in the [`prediction/src/algorithms/classify/trained_model/predict`](https://github.com/concept-to-clinic/concept-to-clinic/blob/5ceab8915d96268d34f638abbcf4b71981cc6d6d/prediction/src/algorithms/classify/trained_model.py#L13) method.
- Code to train the model should live in the [`prediction/src/algorithms/classify/src/`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/prediction/src/algorithms/classify/src) folder
- A fully serialized version of the model that can be loaded should live in the [`prediction/src/algorithms/classify/assets/`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/prediction/src/algorithms/classify/assets) folder using [`git-lfs`](https://git-lfs.github.com/)

## Out of scope

This feature is a first-pass at getting a model that completes the task with the defined input and output. We are not yet judging the model based on its accuracy or computational performance.

## Acceptance criteria

 - [ ] trained model
 - [ ] documentation for the trained model (e.g., cross validation performance, data used) and how to re-train it

*NOTE: All PRs must follow the standard PR checklist.*

reiinakano:
Not sure if this is a dumb question, but given that there will probably be multiple models in the end, how do you intend to separate them? Would it make sense to to further divide `prediction/classify` into a subfolder for each model? e.g. `prediction/classify/vgg` would contain `prediction/classify/vgg/src/`, `prediction/classify/vgg/trained_model/predict/`, and `prediction/classify/vgg/assets/`

Question applicable to #1 and #3

isms:
[...]

PR #122, opened 2017-09-18
Converted Team 'grt123' identification and classification algorithms

dchansen:
## Description

This pull request implements the grt123 identification algorithm.
It is based on PyTorch, which currently conflicts with Tensorflow, so the pull request also disables the Keras import. 

While the algorithm is quite fast on GPU, it is painfully slow (but working) on the CPU due to a bug in PyTorch. The bug is fixed in the PyTorch development branch, but the next release is about 2 months off. 

https://github.com/concept-to-clinic/concept-to-clinic/issues/4
https://github.com/concept-to-clinic/concept-to-clinic/issues/1

## Motivation and Context

This Pull request implements  part of the prediction service (https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#detect-prediction-service) 

## How Has This Been Tested?

The algorithm has largely been tested by manually confirming about 5 random LUNA2016 datasets - as the loaded model has already been tested extensively elsewhere, this was deemed sufficient. 
It has been tested both outside of docker, on a Titan X GPU, and inside docker on an i7 Intel CPU.

## CLA

### DIFF
--- a/prediction/src/algorithms/classify/trained_model.py
+++ b/prediction/src/algorithms/classify/trained_model.py
@@ -7,9 +7,10 @@
     for if nodules are concerning or not.
 """
 
-import numpy as np
-import keras.models
+from src.algorithms.classify.src import gtr123_model
 from src.preprocess.load_ct import load_ct, MetaData
+
+import SimpleITK as sitk
 
 
 def predict(dicom_path, centroids, model_path=None,
@@ -43,23 +44,26 @@ def predict(dicom_path, centroids, model_path=None,
              'z': int,
              'p_concerning': float}
     """
-    if not len(centroids) or model_path is None:
-        return []
+    reader = sitk.ImageSeriesReader()
+    filenames = reader.GetGDCMSeriesFileNames(dicom_path)
 
-    model = keras.models.load_model(model_path)
-    ct_array, meta = load_ct(dicom_path)
-    if preprocess_ct is not None:
-        meta = MetaData(meta)
-        ct_array = preprocess_ct(ct_array, meta)
-        if not isinstance(ct_array, np.ndarray):
-            raise TypeError('The signature of preprocess_ct must be ' +
-                            'callable[list[DICOM], ndarray] -> ndarray')
+    if not filenames:
+        raise ValueError("The path doesn't contain neither .mhd nor .dcm files")
 
-    patches = preprocess_model_input(ct_array, centroids)
-    predictions = model.predict(patches)
-    predictions = predictions.astype(np.float)
+    reader.SetFileNames(filenames)
+    image = reader.Execute()
 
-    for i, centroid in enumerate(centroids):
-        centroid['p_concerning'] = predictions[i, 0]
+    if preprocess_ct:
+        meta = load_ct(dicom_path)[1]
+        voxel_data = preprocess_ct(image, MetaData(meta))
+    else:
+        voxel_data = image
 
-    return centroids
+    if preprocess_model_input:
+        preprocessed = preprocess_model_input(voxel_data, centroids)
+    else:
+        preprocessed = voxel_data
+
+    model_path = model_path or "src/algorithms/classify/assets/gtr123_model.ckpt"
+
+    return gtr123_model.predict(preprocessed, centroids, model_path)
--- a/prediction/src/tests/test_identify_trained_model_predict.py
+++ b/prediction/src/tests/test_identify_trained_model_predict.py
@@ -1,0 +1,49 @@
+import numpy as np
+import pytest
+
+from ..algorithms.identify import trained_model
+from ..tests.test_endpoints import skip_slow_test
+
+
+@pytest.fixture
+def dicom_path_001():
+    yield '../images/LIDC-IDRI-0001/1.3.6.1.4.1.14519.5.2.1.6279.6001.298806137288633453246975630178/' \
+          '1.3.6.1.4.1.14519.5.2.1.6279.6001.179049373636438705059720603192'
+
+
+@pytest.fixture
+def dicom_path_003():
+    yield "/images/LIDC-IDRI-0003/1.3.6.1.4.1.14519.5.2.1.6279.6001.101370605276577556143013894866/" \
+          "1.3.6.1.4.1.14519.5.2.1.6279.6001.170706757615202213033480003264"
+
+
+@pytest.fixture
+def nodule_locations_001():
+    yield {"x": 317, "y": 367, "z": 7}
+
+
+@pytest.fixture
+def nodule_locations_003():
+    yield {"x": 369, "y": 347, "z": 6}
+
+
+@pytest.mark.skipif(skip_slow_test, reason='Takes very long')
+def test_identify_nodules_001(dicom_path_001, nodule_locations_001):
+    predicted = trained_model.predict(dicom_path_001)
+
+    first = predicted[0]
+
+    dist = np.sqrt(np.sum([(first[s] - nodule_locations_001[s]) ** 2 for s in ["x", "y", "z"]]))
+
+    assert (dist < 10)
+
+
+@pytest.mark.skipif(skip_slow_test, reason='Takes very long')
+def test_identify_nodules_003(dicom_path_003, nodule_locations_003):
+    predicted = trained_model.predict(dicom_path_003)
+
+    first = predicted[0]
+
+    dist = np.sqrt(np.sum([(first[s] - nodule_locations_003[s]) ** 2 for s in ["x", "y", "z"]]))
+
+    assert (dist < 10)
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,3 +9,4 @@ test/assets/* filter=lfs diff=lfs merge=lfs -text
 *.hd5 filter=lfs diff=lfs merge=lfs -text
 *.mhd filter=lfs diff=lfs merge=lfs -text
 *.raw filter=lfs diff=lfs merge=lfs -text
+*.ckpt filter=lfs diff=lfs merge=lfs -text
--- a/compose/prediction/Dockerfile-dev
+++ b/compose/prediction/Dockerfile-dev
@@ -1,6 +1,9 @@
-FROM python:3.6
+FROM ubuntu:rolling
 ENV PYTHONUNBUFFERED 1
-
+RUN apt-get update && apt-get install -y tcl tk python3.6 python3.6-tk wget python-opencv
+RUN wget https://bootstrap.pypa.io/get-pip.py
+RUN python3.6 get-pip.py
+RUN ln -s /usr/bin/python3.6 /usr/local/bin/python
 # Requirements have to be pulled and installed here, otherwise caching won't work
 COPY ./prediction/requirements /requirements
 RUN pip install -r /requirements/local.txt
--- a/prediction/requirements/base.txt
+++ b/prediction/requirements/base.txt
@@ -10,3 +10,4 @@ opencv-python==3.3.0.10
 pandas==0.20.3
 scikit-image==0.13.0
 SimpleITK==1.0.1
+http://download.pytorch.org/whl/cu80/torch-0.2.0.post3-cp36-cp36m-manylinux1_x86_64.whl
--- a/prediction/src/algorithms/classify/assets/gtr123_model.ckpt
+++ b/prediction/src/algorithms/classify/assets/gtr123_model.ckpt
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:eb2c037dd55e2f49da95657f7a7851cbcfe2f2b516848ed03f8c5c820f3e16b4
+size 21621745
--- a/prediction/src/algorithms/classify/src/gtr123_model.py
+++ b/prediction/src/algorithms/classify/src/gtr123_model.py
@@ -1,0 +1,315 @@
+import torch
+import numpy as np
+from torch.autograd import Variable
+from torch import nn
+import SimpleITK as sitk
+
+from src.preprocess.gtr123_preprocess import lum_trans, resample
+
+""""
+Classification model from team gtr123
+Code adapted from https://github.com/lfz/DSB2017
+"""
+config = {}
+
+config['crop_size'] = [96, 96, 96]
+config['scaleLim'] = [0.85, 1.15]
+config['radiusLim'] = [6, 100]
+
+config['stride'] = 4
+
+config['detect_th'] = 0.05
+config['conf_th'] = -1
+config['nms_th'] = 0.05
+config['filling_value'] = 160
+
+config['startepoch'] = 20
+config['lr_stage'] = np.array([50, 100, 140, 160])
+config['lr'] = [0.01, 0.001, 0.0001, 0.00001]
+config['miss_ratio'] = 1
+config['miss_thresh'] = 0.03
+config['anchors'] = [10, 30, 60]
+
+
+class PostRes(nn.Module):
+    """ """
+
+    def __init__(self, n_in, n_out, stride=1):
+        super(PostRes, self).__init__()
+        self.conv1 = nn.Conv3d(n_in, n_out, kernel_size=3, stride=stride, padding=1)
+        self.bn1 = nn.BatchNorm3d(n_out)
+        self.relu = nn.ReLU(inplace=True)
+        self.conv2 = nn.Conv3d(n_out, n_out, kernel_size=3, padding=1)
+        self.bn2 = nn.BatchNorm3d(n_out)
+
+        if stride != 1 or n_out != n_in:
+            self.shortcut = nn.Sequential(
+                nn.Conv3d(n_in, n_out, kernel_size=1, stride=stride),
+                nn.BatchNorm3d(n_out))
+        else:
+            self.shortcut = None
+
+    def forward(self, x):
+
+        residual = x
+        if self.shortcut is not None:
+            residual = self.shortcut(x)
+        out = self.conv1(x)
+        out = self.bn1(out)
+        out = self.relu(out)
+        out = self.conv2(out)
+        out = self.bn2(out)
+
+        out += residual
+        out = self.relu(out)
+        return out
+
+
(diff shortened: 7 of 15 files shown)


prediction/src/tests/test_identify_trained_model_predict.py
```diff
@@ -0,0 +1,17 @@
...
+from ..algorithms.identify import trained_model
+
+@pytest.fixture
+def dicom_path():
+    yield '../images/LIDC-IDRI-0001/1.3.6.1.4.1.14519.5.2.1.6279.6001.298806137288633453246975630178/' \
+          '1.3.6.1.4.1.14519.5.2.1.6279.6001.179049373636438705059720603192'
+
+
+
+def test_classify_predict_model_load(dicom_path):
```
WGierke: As @lamby already stated [here](https://github.com/concept-to-clinic/concept-to-clinic/pull/118#issuecomment-330149196) it'd be nice to still have some tests that validate the output of the models. I'm currently working on some tags to only execute these long running tests if one explicitly desires to do so. So feel free to write this test a bit more detailed :)

isms:
> [...] the pull request also disables the Keras import.

@dchansen I'm not sure this is a feasible compromise. Is there a way to reimplement this in Tensorflow?

prediction/src/tests/test_identify_trained_model_predict.py
dchansen: Sure, I'll have a look at that tomorrow.

dchansen:
@isms I agree that it's not a feasible compromise at all. I did it mainly to highlight the issue.

I think we should have a single deep learning framework for the project as a whole, as they don't really play well together. I am not sure if Keras is the best choice, but either way, the project leads should make the decisions. Perhaps we should open an issue?

Converting the model itself into Keras would relatively easy, but converting the weights is somewhat more involved. Converting to Caffe2 or CNTK would be easy, as they all support the ONNX format.

prediction/src/tests/test_identify_trained_model_predict.py
dchansen: I updated the test cases to be a little more comprehensive.

dchansen:
I am a little unsure why, but the conflict between Tensorflow and Pytorch seems gone. It was realted to this issue in Tensorflow https://github.com/tensorflow/models/issues/523

lamby:
This should block on #118 due to the slow test stuff.

reubano:
Looking pretty good! Just a few questions for you. I noticed that in `classify/trained_model/predict`, you commented out the previous code instead of deleted it. Was just curious, why did you decided to keep it around? Also there's still some spacing errors. Can you run `flake8 prediction` in your terminal? Finally, I see this travis error:

```bash
$ git lfs pull
Git LFS: (540 of 540 files, 2 skipped) 497.08 MB / 497.08 MB, 41.20 MB skipped 
[eb2c037dd55e2f49da95657f7a7851cbcfe2f2b516848ed03f8c5c820f3e16b4] Object does not exist on the server: [404] Object does not exist on the server
[3a46182c1c4883e170d5b576d3a2645309f34817c53c16d4f7f29fbdcf799308] Object does not exist on the server: [404] Object does not exist on the server
```

I'll try to pinpoint which files exactly are missing. Meanwhile, can you verify that the model you added to `git-lfs` was properly uploaded? Thanks for your work!

dchansen:
The git lfs issue should be resolved now. 
I kept the previous code because I was unsure what the long-term plans were regarding the style of the project. Conceivably, this pull request could be changed to fit the style of the previous one. 
While I don't agree with all the choices made, it was in many ways nicer than what I contributed here.  The main reason I did not follow it more closely was one of time, as I just wanted a working adaption of the gtr123 model as soon as possible.

I have fixed all the flake8 issues, except the complexity ones. Hopefully, I will have them fixed by the end of the day.

reubano:
> The git lfs issue should be resolved now.

Awesome!

> I kept the previous code because I was unsure what the long-term plans were regarding the style of the project. Conceivably, this pull request could be changed to fit the style of the previous one. While I don't agree with all the choices made, it was in many ways nicer than what I contributed here. The main reason I did not follow it more closely was one of time, as I just wanted a working adaption of the gtr123 model as soon as possible.

Ok, I understand. I'm taking a look to see how to integrate both versions.

> I have fixed all the flake8 issues, except the complexity ones. Hopefully, I will have them fixed by the end of the day.

Great!

reubano:
Ok... here's my attempt. What do you think?

```python
import SimpleITK as sitk
from src.algorithms.classify.src import gtr123_model
from src.preprocess.load_ct import load_ct, MetaData


def predict(dicom_path, centroids, model_path=None,
            preprocess_ct=None, preprocess_model_input=None):
    reader = sitk.ImageSeriesReader()
    filenames = reader.GetGDCMSeriesFileNames(dicom_path)

    if not filenames:
        raise FileNotFoundError()

    reader.SetFileNames(reader.GetGDCMSeriesFileNames(dicom_path))
    image, meta = reader.Execute(), None

    if preprocess_ct:
        meta = meta or load_ct(dicom_path)[1]
        voxel_data = preprocess_ct(image, MetaData(meta))
    else:
        voxel_data = image

    if preprocess_model_input:
        preprocessed = preprocess_model_input(voxel_data, centroids)
    else:
        preprocessed = voxel_data

    return gtr123_model.predict(preprocessed, centroids)
```

Since the keras model is now gone, do you see any reason to continue to keeping `model_path`? If not, I can help you refactor it out.

dchansen:
This would work for now.
I think the long term solution should be to move the gtr123 preprocessing into preprocess_ct. I will probably not have time to do that until next week though.

reubano:
> I think the long term solution should be to move the gtr123 preprocessing into preprocess_ct.

Agreed. However, I think your PR can work without that at this stage (MVP). I looked around some more and saw that `gtr123_model.predict` actually takes a `model_path` param. So the last line could be `return gtr123_model.predict(preprocessed, centroids, model_path)`. I think then, the only change would have to be [#L14-L16 here](https://github.com/concept-to-clinic/concept-to-clinic/blob/44d50164488adb9c84f1b98824854c53ccc12181/prediction/src/tests/test_classify_trained_model_predict.py#L14-L16) in `test_classify_trained_model_predict.py`.

reubano:
Looking good! What are your thoughts about incorporating something like I [mentioned here](https://github.com/concept-to-clinic/concept-to-clinic/pull/122#issuecomment-331403903)? Main reason is just so that we don't disable current functionality with this PR. I think with that, we can get this merged.

dchansen:
I'm trying to do a reasonable merge right now, and algorith.identify.trained_model is giving me a headache.
I think we really need a standardized way of specifying a model, which is independent of frameworks (Keras, Torch, Tensorflow, etc).

I would suggest that each model independently implements a predict method taking only a dicom path (for identification) or a dicom path and a list of candidates. trained_model.predict would then take the same argument, plus a string identifying the algorithm to be used .  How does this sound?

vessemer:
@dchansen I'm right now working on that, also I've tried to convert the grt123 classification model into Keras (TF), but I've faced troubles with the `BatchNorm` layer, despite it transfer weights correctly it feeds me with different results (if I just throw away each `BatchNorm` from the graph then the output becomes just identical, thus I need to dig deeper).
What is related to the way of standardization without transferring the model I've almost done.

dchansen:
@vessemer I am a little unsure about the benefit of moving the model to Keras now that Tensorflow and Torch seem to work without disturbing one another. Different approaches will require different enough post- and pre-processing  that they can't be represented as a Keras model anyway. 
Long term, I do see the advantage of having all models on a single deep-learning framework, but is Keras the right choice? Using [ONNX](https://github.com/onnx/onnx) as the way to save the model would allow for easy use of many frameworks, including Caffe2 and CNTK. 
Either way, as soon as there is a consensus, I would be happy to help converting the models to that particular format.

One difference between the Torch and Keras batchnorm is that they use a different epsilon value (1e-5 vs 1e-3), so maybe that's the issue? 
Could you make your WIP available on the model standardization?

dchansen:
@reubano Yes, that was my thought exactly. The problem right now is that all the models take ownership of trained_model.predict, and there is no good way of switching between particular models. Ideally, we might also like to do meta-model, combining many different ones, and a uniform interface would help greatly.

vessemer:
@dchansen, thank you for your input, but I've taken into account the difference of the default values, I've used the [nn-transfer](https://github.com/gzuidhof/nn-transfer) for the model standardization.

dchansen:
@vessemer Well, if you can share the code, I can have a look at it.

reubano:
> The problem right now is that all the models take ownership of trained_model.predict, and there is no good way of switching between particular models. Ideally, we might also like to do meta-model, combining many different ones, and a uniform interface would help greatly.

For now, I think this is the preferred course of action. We want to focus on one main model for each of (classify, identify, and segment). Later on, we may consider making it easier to switch out the models. The current area of flexibility is in the serialized trained data, e.g., `assets/dsb2017_detector.ckpt`. So while it is still tied to the underlying python ML model, you can tweak the training parameters as needed. See [my comment on #2](https://github.com/concept-to-clinic/concept-to-clinic/issues/2#issuecomment-325328384) for more info.

Also, let me know what exactly is giving you a headache so I can try and help you out.

dchansen:
My issue was that I had to disable a lot of code  in[identify.trained_model](https://github.com/dchansen/concept-to-clinic/blob/3396244f45f6683bdcc74b8f05081c16549ae315/prediction/src/algorithms/identify/trained_model.py#L22-L71)
But I think this is what you want?

reubano:
I was literally suggesting you just copy/pasted [my code from here](https://github.com/concept-to-clinic/concept-to-clinic/pull/122#issuecomment-331403903).

prediction/src/algorithms/classify/trained_model.py
```diff
@@ -43,23 +43,24 @@ def predict(dicom_path, centroids, model_path=None,
...
+    else:
+        voxel_data = image
 
-    return centroids
+    if preprocess_model_input:
+        preprocessed = preprocess_model_input(voxel_data, centroids)
+    else:
+        preprocessed = voxel_data
+
+    return gtr123_model.predict(preprocessed, centroids)
```
reubano:
```python
...
model_path = model or 'src/algorithms/classify/assets/gtr123_model.ckpt'
...
return gtr123_model.predict(image, centroids, model_path)
```

reubano:
Just about there! Does this solve your headache with `algorithms.identify.trained_model`?

prediction/src/algorithms/classify/trained_model.py
dchansen: Not sure what you're asking here?

reubano: Just reintroducing the `model_path` part [from here](https://github.com/concept-to-clinic/concept-to-clinic/pull/122/commits/ddcb4528777c4c097de823dad86f8708397fa1fe)

reubano:
From the travis log

```bash
$ flake8 prediction
prediction/src/algorithms/identify/trained_model.py:13:1: F401 'glob' imported but unused
prediction/src/algorithms/identify/trained_model.py:15:1: F401 'dicom' imported but unused
prediction/src/algorithms/identify/trained_model.py:16:1: F401 'src.preprocess.errors.EmptyDicomSeriesException' imported but unused
prediction/src/algorithms/identify/trained_model.py:17:1: F401 'src.preprocess.lung_segmentation.save_lung_segments' imported but unused
prediction/src/algorithms/classify/trained_model.py:68:56: E231 missing whitespace after ','
prediction/src/preprocess/extract_lungs.py:57:19: E127 continuation line over-indented for visual indent
prediction/src/preprocess/extract_lungs.py:66:18: E127 continuation line over-indented for visual indent
prediction/src/preprocess/extract_lungs.py:67:18: E127 continuation line over-indented for visual indent
prediction/src/preprocess/extract_lungs.py:68:18: E127 continuation line over-indented for visual indent
prediction/src/tests/test_endpoints.py:10:1: F811 redefinition of unused 'json' from line 7
The command "flake8 prediction" exited with 1.

$ pycodestyle prediction
prediction/src/algorithms/classify/trained_model.py:68:56: E231 missing whitespace after ','
prediction/src/preprocess/extract_lungs.py:57:19: E127 continuation line over-indented for visual indent
prediction/src/preprocess/extract_lungs.py:66:18: E127 continuation line over-indented for visual indent
prediction/src/preprocess/extract_lungs.py:67:18: E127 continuation line over-indented for visual indent
prediction/src/preprocess/extract_lungs.py:68:18: E127 continuation line over-indented for visual indent
```

lamby:
@dchansen This now has conflicts with ``master`` — could you rebase? :)

prediction/src/tests/test_identify_trained_model_predict.py
```diff
@@ -0,0 +1,57 @@
+import os
+
+import numpy as np
+import pytest
+
+from ..algorithms.identify import trained_model
+
+
+def skip_slow_test():
```
WGierke: Sorry for giving feedback at such a late state but you're not supposed to re-implement this function as it was already added to [prediction/src/tests/test_endpoints.py](https://github.com/concept-to-clinic/concept-to-clinic/blob/757dfea9959db8e79f9b4ab5c04748aea711ca77/prediction/src/tests/test_endpoints.py) :) Are you able to import it from there?

dchansen:
@lamby Github claims that it has no conflicts with the base branch?

reubano:
Still some failing tests... travis doesn't show details however, so I'm looking into it locally.

prediction/src/tests/test_identify_trained_model_predict.py
reubano: nice catch @WGierke!

dchansen:
The test run in my local environment (outside docker). Inside docker they run when individually, but not when running them all at once. I suspect it is tensorflow/torch shenanigans, but it is quite tricky to debug.

dchansen:
Using a different docker image solved the problem for me.
I cannot for the life of me figure out where the segfault is coming from in the python3:6 image however.

reubano:
Yea that is odd... which image works? And what is the one that doesn't?

dchansen:
`pytorch:36` segfaults. 
`varunagrawal/pytorch` does not, but I had to manually install opencv and tkinter in it.

dchansen:
I changed the image to be Ubuntu based instead. All tests should pass now.

reubano:
Ok.... looks like the tests passed, but there are some new lint errors now

```bash
$ flake8 prediction
prediction/src/tests/__init__.py:1:1: F401 'torch' imported but unused
prediction/src/tests/__init__.py:1:13: E261 at least two spaces before inline comment
prediction/src/tests/__init__.py:1:14: E262 inline comment should start with '# '
The command "flake8 prediction" exited with 1.

$ pycodestyle prediction
prediction/src/tests/__init__.py:1:13: E261 at least two spaces before inline comment
prediction/src/tests/__init__.py:1:14: E262 inline comment should start with '# '
The command "pycodestyle prediction" exited with 1.
```

dchansen:
The torch import is required simply because torch needs to be imported before Keras, always. I am very open to other ways of doing that.

reubano:
Try adding this on the offending line: `# noqa # pylint: disable=unused-import`

reubano:
Looks like everything is a go now! Do you mind squashing to 1 commit?

dchansen:
I seem to have messed that up rather thoroughly. This will take me a while to sort out.

reubano:
No worries :). Perhaps I can help out. What seems to be the issue(s)?

dchansen:
Just not very experienced in using git rebase. I found myself in the situation of having to redo every merge made on the project for the last ten days. 
I have messed up my repository quite badly - I think my best approach is to try again from home, where I have a pristine version of my repo, before my attempts at rebasing.

WGierke:
Instead of rebasing, I prefer to squash the commits like [so](https://gist.github.com/patik/b8a9dc5cd356f9f6f980). It fulfills the purpose as well, namely having only 1 commit in the end :)

reubano:
You can also take a look at the [advice I posted here](https://github.com/concept-to-clinic/concept-to-clinic/pull/55#issuecomment-325956598).

reubano:
Also `git config --global rerere.enabled true` will save conflict resolutions so in that in the future, you only have to resolve them once.

dchansen:
@WGierke Thank you! Uploading the squashed commit now.

reubano:
```bash
$ flake8 interface
interface/backend/api/tests.py:44:1: W293 blank line contains whitespace
interface/backend/api/tests.py:48:57: E502 the backslash is redundant between brackets
interface/backend/api/tests.py:49:81: E502 the backslash is redundant between brackets
interface/backend/api/serializers.py:74:1: E302 expected 2 blank lines, found 1
interface/backend/api/serializers.py:120:17: E127 continuation line over-indented for visual indent
interface/backend/api/views.py:24:1: F811 redefinition of unused 'Response' from line 13
interface/backend/api/views.py:47:37: W291 trailing whitespace
The command "flake8 interface" exited with 1.

$ pycodestyle interface
interface/backend/api/serializers.py:74:1: E302 expected 2 blank lines, found 1
interface/backend/api/serializers.py:120:17: E127 continuation line over-indented for visual indent
interface/backend/api/tests.py:44:1: W293 blank line contains whitespace
interface/backend/api/tests.py:48:57: E502 the backslash is redundant between brackets
interface/backend/api/tests.py:49:81: E502 the backslash is redundant between brackets
interface/backend/api/views.py:47:37: W291 trailing whitespace
The command "pycodestyle interface" exited with 1.
```

dchansen:
But the pull request doesn't include those files??

isms:
@reubano @dchansen Those issues were introduced by recent commits unrelated to this PR. I pushed fixes which should get the code style back to passing. Will restart the build on this branch to update the status.

(cc @lamby so we can avoid in the future!)

reubano:
Ok... master looks good. You mind rebasing @dchansen to ensure everything is good to go?

reubano:
closes #1 and #4 :shipit:

Issue #110, opened 2017-09-12
Port existing pages to frontend Vue project


antonow:
Before the base Vue app was merged ([PR 97](https://github.com/concept-to-clinic/concept-to-clinic/pull/97)), several frontend pages were developed and merged. These pages were being served from Django. It was later decided that all pages should be served from the single instance of the Vue app (see [here](https://github.com/concept-to-clinic/concept-to-clinic/issues/106#issuecomment-328642670)). As a result, these existing pages need to be ported over to the Vue app.

## Expected Behavior
The base Vue app in `interface/frontend/src/main.js` should serve all existing page templates using `VueRouter`. Clicking on a link should load the expected page template.

## Current Behavior
At the moment, only the navigation is displayed when loading the home route from Django and clicking on links in the navigation does not render any of the existing page templates.

## Possible Solution
The code from PR's [73](https://github.com/concept-to-clinic/concept-to-clinic/pull/73), [84](https://github.com/concept-to-clinic/concept-to-clinic/pull/84), [85](https://github.com/concept-to-clinic/concept-to-clinic/pull/85), [88](https://github.com/concept-to-clinic/concept-to-clinic/pull/88), [92](https://github.com/concept-to-clinic/concept-to-clinic/pull/92), and [93](https://github.com/concept-to-clinic/concept-to-clinic/pull/93) will need to be modified to use ES6 syntax and served from the Vue app using `VueRouter`.

## Steps to Reproduce
1. Fix the issue with the broken asset link as described in [issue 106](https://github.com/concept-to-clinic/concept-to-clinic/issues/106)
2. `cd interface/frontend && npm install && npm run build`
3. `cd ../..`
4. `docker-compose -f local.yml up`
5. Visit `localhost:8000` -- only the navigation appears

## Context (Environment)
Until the existing work is integrated into the base Vue app, it will be difficult to develop the remaining features which need to be built on top of these existing templates.

## Detailed Description
I am proposing to set up `VueRouter` and add all the necessary frontend routes to serve the page templates that have already been developed. These existing Vue components will also need to be updated to ES6 syntax in order for the frontend build process to succeed.

## Checklist before submitting
- [x] I have confirmed this using the officially supported Docker Compose setup using the `local.py` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [X] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
[...]

PR #125, opened 2017-09-19
Port existing views and add VueRouter and Axios

antonow:
## Description
This PR takes the existing views built by @WGierke, including his unmerged frontend work on [#34 Select Lung Orientation](https://github.com/concept-to-clinic/concept-to-clinic/pull/101), and turns them into components of the main Vue app. VueRouter was set up to handle page navigation, and jQuery and Bootstrap.js were removed. Axios, an alternative to jQuery's Ajax, was added because it makes it easier to set default request options. I also made a few backend changes because I was getting errors when trying to create some test data through the Django REST interface.

[Port existing pages to frontend Vue project #110](https://github.com/concept-to-clinic/concept-to-clinic/issues/110)

## Motivation and Context
This change will allow us to serve all templates from the main Vue app, and sets up basic routing and frontend file organization patterns that will give the app some structure moving forward.

## How Has This Been Tested?
This has been tested by running `docker-compose -f local.yml build` to install new dependencies, followed by `docker-compose -f local.yml up`, and navigating to all the links in the navigation. I created test data using the Django REST interface to verify that actions like toggling elements that can be opened and closed and making backend requests works as expected.

screen shot 2017-09-18 at 8 27 20 pm
screen shot 2017-09-18 at 8 27 58 pm
screen shot 2017-09-18 at 8 28 20 pm

## CLA


interface/frontend/src/components/Nodule.vue
```diff
@@ -0,0 +1,62 @@
...
+  },
+  created () {
+    if (this.index === 0) { this.isOpen = true }
+  },
+  methods: {
+    toggleShow (nodule) {
+      this.isOpen = !this.isOpen
+    },
+    update (nodule) {
+      this.$axios.put(nodule.url, { lung_orientation: this.selected }).then((response) => {
```
antonow: FYI, at the moment `PUT` requests are failing because of [Preflighted requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests). They are causing an `OPTION` request to occur which fails likely because the backend API isn't set up to respond to that type of request. I believe the solution is to allow CORS on the backend.

antonow: Ok correction: looks like `OPTIONS` requests are actually allowed, but `PUT` is not.

screen shot 2017-09-18 at 8 55 55 pm

WGierke:
Great stuff! If you add one more empty line to `serializers.py:38`, the Travis tests should pass :)
```
(venv3.5) ➜  concept-to-clinic_old git:(71855eb) flake8 interface && flake8 prediction
interface/backend/api/serializers.py:38:1: E302 expected 2 blank lines, found 1
```

lamby:
Is this ready to roll? I see some unaddressed comments and/or commentary re. HTTP methods above? :)

antonow:
Yeah, I was just pointing out that the API endpoints for some of these frontend actions aren't in place yet. I don't think adding those endpoints is within the scope of this issue though.

interface/frontend/src/components/CandidateList.vue
```diff
@@ -0,0 +1,47 @@
...
+    return {
+      candidates: []
+    }
+  },
+  created () {
+    this.fetchCandidates()
+  },
+  methods: {
+    fetchCandidates () {
+      this.$axios.get('/api/candidates').then(
```
lamby: This should be a ``{% url "..." %}``?

antonow: @lamby This won't work when developing with the dev server because the files are being served by the dev server instead of being rendered by Django. But it also wouldn't work with our current production set-up because the [index file](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/interface/frontend/src/index.html) that gets served from Django only contains an empty `<app>` component into which the remainder of the frontend app gets rendered via JS. We can pass information from Django into the app by adding data as properties of the `<app>` component, but it seems a bit of a hack to me.

lamby: Can you clarify this in the code itself? Otherwise this comment will be lost.. :)

antonow: @lamby, added this comment to the [index file](https://github.com/concept-to-clinic/concept-to-clinic/pull/125/files#diff-f0219d0aabd1326ced417e7bd31ebc7d).

antonow:
FYI, I added a couple small improvements to this PR since this was last reviewed.

lamby:
thanks!

Issue #120, opened 2017-09-18
Lung Segmentation


vessemer:
Prior to stages [#1](https://github.com/concept-to-clinic/concept-to-clinic/issues/1), [#2](https://github.com/concept-to-clinic/concept-to-clinic/issues/2), and [#3](https://github.com/concept-to-clinic/concept-to-clinic/issues/3), CT scans must be preprocessed. Preprocessing involves various steps such as clipping by value, rescaling, and magnitude normalization. The fact that many of the top solutions contain these steps in their pipelines, highlights their importance. 

There exists various approaches dedicated to lungs segmentation e.g. via [Hounsfield value](https://www.kaggle.com/gzuidhof/full-preprocessing-tutorial), affine/rigid/non-rigid [image registration](http://simpleelastix.readthedocs.io/GettingStarted.html), [watershed algorithm](https://www.kaggle.com/ankasor/improved-lung-segmentation-using-watershed) etc. The tradeoff these approaches usually make comes down to amount of computing resources consuming vs quality of segmentation.

## Expected Behavior
Adopted or created lung segmentation algorithm should lie in the `src/preprocess/lungs_segmentation.py`. It also should be integrated into classes `src.preprocess.preprocess_ct.PreprocessCT` and `src.preprocess.preprocess_ct.Params`.

## Current Behavior
For now `PreprocessDicom` doesn't contain any lung segmentation algorithm.

## Acceptance criteria
- [ ] method aimed at lungs segmentation
- [ ] brief justification of a selected approach
- [ ] tests for the methods

reubano:
Thanks for this @vessemer. The links you provided were very informative! I learned quite a bit just gleaming through them. My initial thought is, how much of this should be handled by #3 vs common functions that would be shared across #1, #2, and #3?

vessemer:
@reubano, I guess you're right, it'll be convenient to place the lung segmentation method inside the `segment` directory. For now, the project has been structured in such a way that all `algorithm`'s subdirectories contain `trained_model.py`. Perhaps, it's worth to reform the `segment` part, so that latter will contain `lungs`, `nodules`  or something similar. Since all stages performed prior to the classification may be roughly counted as preprocessing.

reubano:
> Perhaps, it's worth to reform the segment part, so that latter will contain lungs, nodules or something similar. 

[...]

PR #132, opened 2017-09-21
Data Generator

vessemer:
The data augmentation process is a very common stage for a plenty of approaches in machine learning, especially in computer vision. The majority of the solutions of the #1, #2, #3 has the similar augmentation pipeline (e.g. [grt123](https://github.com/WGierke/concept-to-clinic/blob/457b7a7528fda28ed556d234135d48595eaa13ce/docs/algorithm_grt123.md), [Therapixel](https://github.com/WGierke/concept-to-clinic/blob/d9a20abfb817642ba6d98e403ed6659446125853/docs/algorithm_therapixel.md), [Daniel Hammack](https://github.com/WGierke/concept-to-clinic/blob/a47d8f28435f612d93a8eee694ac5c82500e710f/docs/algorithm_hammack.md), etc.). The aforementioned pipeline mainly consists of Hounsfield scaling, lungs segmentation (#120), CT re-orientation and a batch of trivial, yet resource consumptive spacial operations: zoom, rotation, shear, shift, flip and combinations of them. The convenient solution for the latter process is to build a generator which will yield a processed patches via affine transformations. The worth notable example of such a generator is [Keras DataGenerator](https://keras.io/preprocessing/image/) complemented by the [thorough description](https://blog.keras.io/building-powerful-image-classification-models-using-very-little-data.html). The main advantage of the Keras approach is that it for each patch build new transformation matrix from scratch but apply it only ones in comparison with step-by-step application of `scipy.ndimage` perturbations e.g. [code of Daniel Hammack](https://github.com/dhammack/DSB2017/blob/67f6474768fa47f9be95c31aa41609e0a986d8a2/training_code/aws/data_generator_fn3.py#L126-L128) where only for a rotation around the origin two calls of affine transformations were performed. Unfortunately, the main disadvantage for us is that Keras DataGenerator doesn't work with 3D data. Thus what have I done is the extension of the DataGenerator in order to deal with volumetric data such as CT scans. I've also achieved compatibility with `scipy.ndimage` to check integrity of the methods by the trusted functions (i.e. `scipy.ndimage.zoom`, `.rotate`, etc.) which I further have used in the unit tests. Resulted speedup is about 2.5 in comparison with sequantial calls of `scipy.ndimage` functions.

### Random Rotation
rand_rotation_b
rand_rotation_full_b

### Random Zoom 
rand_zoom_b

### Random Shear 
random_shear

### Random Shift
rand_shift_b

### Agregated
one
two

## CLA

### DIFF
--- a/prediction/src/preprocess/crop_patches.py
+++ b/prediction/src/preprocess/crop_patches.py
@@ -1,0 +1,77 @@
+import itertools
+
+import numpy as np
+import scipy.ndimage
+
+
+def mm2voxel(coord, origin=0., spacing=1.):
+    """ Transfer coordinates in mm into voxel's location
+
+    Args:
+        coord (scalar | list[scalar]): coordinates in mm.
+        origin (scalar | list[scalar]): an origin of the CT scan in mm.
+        spacing (scalar | list[scalar]): an CT scan's spacing, i.e. the size of one voxel in mm.
+
+    Returns:
+        list[int]: the voxel location related to the coord.
+    """
+    if np.isscalar(coord):
+        coord = [coord]
+    coord = np.array(coord)
+    origin = scipy.ndimage._ni_support._normalize_sequence(origin, len(coord))
+    spacing = scipy.ndimage._ni_support._normalize_sequence(spacing, len(coord))
+    coord = np.ceil(coord - np.array(origin)) / np.array(spacing)
+    return coord.astype(np.int).tolist()
+
+
+def crop_patch(ct_array, patch_shape, centroids, meta):
+    """ Generator yield a patch of a desired shape for each centroid
+    from a given a CT scan.
+
+    Args:
+        ct_array (np.ndarray): a numpy ndarray representation of a CT scan
+        patch_shape (int, list[int]): a desired shape of a patch. If int will be provided,
+            then patch will be a cube-shaped.
+        centroids (list[dict]): A list of centroids of the form::
+            {'x': int,
+             'y': int,
+             'z': int}
+        meta (src.preprocess.load_ct.MetaData): meta information of the CT scan.
+
+    Yields:
+        np.ndarray: a cropped patch from the CT scan.
+    """
+    patch_shape = scipy.ndimage._ni_support._normalize_sequence(patch_shape, len(ct_array.shape))
+    padding = np.ceil(np.array(patch_shape) / 2.).astype(np.int)
+    padding = np.stack([padding, padding], axis=1)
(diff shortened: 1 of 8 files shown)


prediction/src/tests/test_load_dicom.py
```diff
@@ -5,8 +5,8 @@
 import numpy as np
 import pytest
 
-from ..preprocess import errors
```
lamby: How come?

prediction/src/preprocess/preprocess_ct.py
```diff
@@ -82,7 +82,9 @@ def __init__(self, params=None):
 
     def __call__(self, voxel_data, meta):
         if not isinstance(meta, load_ct.MetaData):
-            raise ValueError('The meta should be an instance of %s.' % str(load_ct.MetaData))
+            print(meta.__class__)
```
lamby: Whoops? :)

prediction/src/preprocess/load_ct.py
```diff
@@ -127,14 +127,26 @@ def extract_spacing_dcm(self):
...
+            return [spacing[1], spacing[0], slice_thickness]
+        else:
+            return [slice_thickness, spacing[0], spacing[1]]
 
     def extract_spacing_mhd(self):
-        # Again, taking into account ijk -> xyz transformation
-        return list(reversed(self.meta.GetSpacing()))
+        if self.xyz_order:
+            return list(reversed(self.meta.GetSpacing()))
+        else:
```
lamby: No need for an else; you've just called ``return`` on the previous lin. (and elsewhere)

prediction/src/preprocess/crop_patches.py
```diff
@@ -0,0 +1,80 @@
...
+    Yields:
+        np.ndarray: a cropped patch from the CT scan.
+    """
+    patch = crop_patch(ct_array, patch_shape, centroids, meta)
+    patches = list()
+    while True:
+        try:
+            patches.append(next(patch))
+        except StopIteration:
+
```
lamby: flake might complain here, not sure...

vessemer: Looks ugly, agreed. I've replaced it via `itertools.islice`.

vessemer:
@lamby, all rebased and stashed.

reubano:
I'm having trouble figuring out what issue this addresses. Ideas @vessemer @lamby?

reubano:
Created #137 to accompany this

PR #133, opened 2017-09-21
#120 Lung Segmentation (Hounsfield based)

WGierke:
I refactored the current identification algorithm that's based on the approach of [Julian de Wit](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/docs/algorithm_de_wit.md) which uses Hounsfield values between -1000 and 400 to segment lungs.
As a byproduct, the algorithm already saves images of pre-processed scans and the related lung masks. I refactored the structure so the segmentation method can be accessed easier by other algorithms as well.

This addresses #120 .

## Motivation and Context
Every nodule identification algorithm is supposed to somehow segment the lung first. Thus, it makes sense to have one separate function that is shared among the different identification algorithms.

## How Has This Been Tested?
I ran the [nodule identification test](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/prediction/src/tests/test_endpoints.py#L76) which saved files of the segmented lungs as PNG files to `data/extracted/USER_ID/`. I constructed the following GIF based on them, which shows how well the lung got segmented.
result1
I'm very open for suggestions for more "code-heavy" tests ;)

## CLA


prediction/src/preprocess/lung_segmentation.py
```diff
@@ -0,0 +1,142 @@
...
+        dicom_path: a path to a DICOM directory
+        patient_id: SeriesInstanceUID of the patient
+
+    Returns:
+        shape of the patient images (z, x, y)
+    """
+    EXTRACTED_IMAGE_DIR = "data/extracted/"
+    TARGET_VOXEL_MM = 1.00
+
+    target_dir = EXTRACTED_IMAGE_DIR + patient_id + "/"
```
lamby: ``os.path.join``??

prediction/src/preprocess/lung_segmentation.py
```diff
@@ -0,0 +1,142 @@
...
+    pixel_spacing = slices[0].PixelSpacing
+    pixel_spacing.append(slices[0].SliceThickness)
+    image = rescale_patient_images(image, pixel_spacing, TARGET_VOXEL_MM)
+    if not invert_order:
+        image = numpy.flipud(image)
+
+    for i in range(image.shape[0]):
+        patient_dir = target_dir
+        if not os.path.exists(patient_dir):
+            os.mkdir(patient_dir)
```
lamby: I think there is ``mkdirs`` with some kind of ``ignore`` flag? :)

prediction/src/preprocess/lung_segmentation.py
```diff
@@ -0,0 +1,142 @@
...
+    image = get_pixels_hu(slices)
+    invert_order = slices[1].ImagePositionPatient[2] > slices[0].ImagePositionPatient[2]
+
+    pixel_spacing = slices[0].PixelSpacing
+    pixel_spacing.append(slices[0].SliceThickness)
+    image = rescale_patient_images(image, pixel_spacing, TARGET_VOXEL_MM)
+    if not invert_order:
+        image = numpy.flipud(image)
+
+    for i in range(image.shape[0]):
```
lamby: Typically one uses ``enumerate`` instead

prediction/src/preprocess/lung_segmentation.py
```diff
@@ -0,0 +1,142 @@
...
+        cv2.imwrite(img_path.replace("_i.png", "_m.png"), mask * 255)
+
+    return image.shape
+
+
+def load_patient(src_dir):
+    slices = []
+    for s in os.listdir(src_dir):
+        try:
+            slices.append(dicom.read_file(src_dir + '/' + s))
```
lamby: ``os.path.join`` etc

reubano:
whoa... that gif is awesome! ✨

isms:
@WGierke Simple (dumb?) idea for testing accuracy here: what if you used this segmentation to create a mask for every single image in the LIDC-IDRI data and then looked at what % of identified nodules fell inside the mask vs. outside the mask?

NB, this wouldn't be part of the ordinary test suite because it assumes that all the images are available but it would be helpful to know *for science* :microscope:

WGierke:
@isms Thanks for your input. If I understand you correctly, the issue with the current identification algorithm by De Wit will be that it only considers tissue that is inside this mask, so it's not possible at all to find nodules outside the lung. I could rewrite the identification algorithm to consider the whole image and not just the lung but what would we expect to happen in that case? Only detecting nodules in an area that corresponds to the lung would be nice. However, if the algorithm detects nodules in areas that do not belong to the lung, the reason could also be that the model simply wasn't trained to distinguish between lung and not lung on its own - so the model could also be "false".

isms:
@WGierke Actually, I was only remarking that _if_ the goal is to see how well this segments lung vs not lung, and we assume that all labeled lung nodules are actually inside the lungs, seeing how many of the labeled nodules are inside these segments would tell us more about how good the lung/not-lung segmentation is.

Does that make sense?

WGierke:
@isms Oh alright, sorry, now I got it. Yes, thank definitely makes sense. I'll add that :)

WGierke:
@isms I added a [test](https://github.com/concept-to-clinic/concept-to-clinic/pull/133/files#diff-e6c519f70a6bee9cab764b751df70348) that checks whether the annotations of the local LIDC images are inside the segmented lung masks. I'm using [pylidc](https://github.com/pylidc/pylidc) for that which makes it easier to query attributes of the LIDC images (like annotations). However, the library has to be installed after the other requirements, otherwise it throws some `ModuleNotFoundError`s.

isms:
@WGierke Excellent, thanks for the footwork there.

Pending @lamby's satisfaction with the requested changes, I'm in favor of merging :+1:

prediction/src/preprocess/lung_segmentation.py
```diff
@@ -0,0 +1,161 @@
...
+
+    return original_image.shape, image.shape
+
+
+def load_patient(src_dir):
+    slices = []
+    for s in os.listdir(src_dir):
+        try:
+            slices.append(dicom.read_file(os.path.join(src_dir, s)))
+        except Exception as e:
```
reubano: This is the only that really caught my eye. Specific exceptions are usually best. But in this case, the error is being logged, so it isn't that big of a deal. If you twisted my arm, I'd probably say something like this may be a bit cleaner (if `SpecificException` is in fact the only expected exception).

```
try:
    slice = dicom.read_file(os.path.join(src_dir, s))
except SpecificException:
    logging.error("{} is not a valid DICOM".format(s))
else:
    slices.append(slice)
```

WGierke: Good point, thanks! I replaced the twisted code by your suggestion ;)

lamby:
LGTM :)

PR #142, opened 2017-09-29
Implemented simple nodule segmentation algorithm

dchansen:
This pull request implements a simple level-set based segmentation algorithm for nodule segmentation.

## Description

This implements a simple segmentation algorithm based on SITK, and returns binary mask, diameter and volume.

#3 

## Motivation and Context

Part of minimal viable product.

## How Has This Been Tested?

The segmentation has been tested manually on a few LIDC images. A better approach would be a test based on pylidc and report dice values.

Nodule Segmentation

## CLA

### DIFF
--- a/prediction/src/algorithms/segment/src/levelset_segment.py
+++ b/prediction/src/algorithms/segment/src/levelset_segment.py
@@ -1,0 +1,63 @@
+import SimpleITK as sitk
+
+
+def segment_nodule(image: sitk.Image, nodule: dict, lower_threshold=-600, upper_threshold=500, maximum_rms_error=0.02,
+                   number_of_iterations=1000, curvature_scaling=0.5, radius=3):
+    """
+    Given an image and a nodule location, this function segments the nodule, using the
+    ThresholdSegmentationLevelSetImageFilter from ITK.
+    See https://itk.org/Doxygen/html/classitk_1_1SegmentationLevelSetImageFilter.html for further details on the Method
+
+    Args:
+      image: sitk.Image: Input Image
+      nodule: dict: Nodule coordinates, in the form {"x": 1, "y": 1, "z": 1}
+      lower_threshold:  Lower Hu threshold of the nodule to be segmented. Should be about the Hu of lung
+                      (Default value = -600)
+      upper_threshold: Upper Hu threshold of the nodule to be segmented. Should be less than the Hu of bone
+                       (Default value = 500)
+      maximum_rms_error: Stopping condition for the iterative solver.
+      number_of_iterations: Maximum number of iterations to run
+      curvature_scaling: Controls the smoothness of the contour. Higher value - smoother contour
+      radius: Sets the radius of the initial segmentation around the nodule.
+
+    Returns:
+        mask: filepath to segmentation mask
+        diameters: list of diameters
+        volumes: list of  volumes
+
+    """
+
+    seed = (nodule["x"], nodule["y"], nodule["z"])
+    # Create initial levelset based on the seed and the signed Maurer distance
+    seg = sitk.Image(image.GetSize(), sitk.sitkUInt8)
+    seg.CopyInformation(image)
+    seg[seed] = 1
+    # Binary dilate enlarges the seed mask by 3 pixels in all directions.
+    seg = sitk.BinaryDilate(seg, radius)
+    init_ls = sitk.SignedMaurerDistanceMap(seg, insideIsPositive=True, useImageSpacing=True)
+
+    lsFilter = sitk.ThresholdSegmentationLevelSetImageFilter()
+    # Sets the Hu value range for the nodule
+    lsFilter.SetLowerThreshold(lower_threshold)
+    lsFilter.SetUpperThreshold(upper_threshold)
+    # Set stopping conditions
+    lsFilter.SetMaximumRMSError(maximum_rms_error)
+    lsFilter.SetNumberOfIterations(number_of_iterations)
+    # CurvatureScaling controls how smooth the contour will be.
+    lsFilter.SetCurvatureScaling(curvature_scaling)
+    # PropagationScaling is also known as "balloon-force". Causes the segmented area to grow.
+    lsFilter.SetPropagationScaling(1)
+    lsFilter.ReverseExpansionDirectionOn()
+    ls = lsFilter.Execute(init_ls, sitk.Cast(image, sitk.sitkFloat32))
+    # Convert to binary mask
+    mask = ls > 0
+
+    # Calculate diameter and volume
+    stats = sitk.LabelShapeStatisticsImageFilter()
+    stats.ComputeFeretDiameterOn()
+    stats.SetBackgroundValue(0)
+    stats.Execute(mask)
+    diameter = stats.GetFeretDiameter(1)
+    volume = stats.GetPhysicalSize(1)
+
+    return mask, diameter, volume
--- a/prediction/src/algorithms/segment/trained_model.py
+++ b/prediction/src/algorithms/segment/trained_model.py
@@ -8,10 +8,11 @@
 """
 
 from src.preprocess.load_ct import load_ct, MetaData
-
+from src.algorithms.segment.src import levelset_segment
+from src import utils
 import numpy as np
-import os
 import scipy.ndimage
+import SimpleITK as sitk
 
 
 def predict(dicom_path, centroids):
@@ -36,15 +37,30 @@ def predict(dicom_path, centroids):
         dict: Dictionary containing path to serialized binary masks and
             volumes per centroid with form::
             {'binary_mask_path': str,
-             'volumes': list[float]}
+             'volumes': list[float]
+             'diameters': list[float]}
     """
-    load_ct(dicom_path)
-    segment_path = os.path.join(os.path.dirname(__file__),
-                                'assets', 'test_mask.npy')
-    volumes = calculate_volume(segment_path, centroids)
+
+    reader = sitk.ImageSeriesReader()
+    reader.SetFileNames(reader.GetGDCMSeriesFileNames(dicom_path))
+    image = reader.Execute()
+
+    binary_mask = np.zeros_like(sitk.GetArrayFromImage(image), dtype=bool)
+
+    volumes = []
+    diameters = []
+    for nodule in centroids:
+        mask, diameter, volume = levelset_segment.segment_nodule(image, nodule)
+        volumes.append(volume)
+        diameters.append(diameter)
+        binary_mask |= sitk.GetArrayFromImage(mask).astype(bool)
+
+    segment_path = utils.get_temporary_file(".npy")
+    np.save(segment_path, binary_mask)
     return_value = {
-        'binary_mask_path': segment_path,
-        'volumes': volumes
+        'binary_mask_path': segment_path.name,
+        'volumes': volumes,
+        'diameters': diameters
     }
 
     return return_value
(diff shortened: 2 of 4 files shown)


prediction/src/algorithms/segment/src/levelset_segment.py
```diff
@@ -0,0 +1,37 @@
...
+
+    seed = (nodule["x"], nodule["y"], nodule["z"])
+    seg = sitk.Image(image.GetSize(), sitk.sitkUInt8)
+    seg.CopyInformation(image)
+    seg[seed] = 1
+    seg = sitk.BinaryDilate(seg, 3)
+
+    init_ls = sitk.SignedMaurerDistanceMap(seg, insideIsPositive=True, useImageSpacing=True)
+    lsFilter = sitk.ThresholdSegmentationLevelSetImageFilter()
+    lsFilter.SetLowerThreshold(-600)
```
WGierke: Could you explain these magic numbers in meaningful constant variables, please?

WGierke:
Nice stuff! Do you have some measurement for the accuracy of your segmentation algorithm on the LIDC dataset? That would make it easier to compare it against further approaches :)

dchansen:
@WGierke This was very much a minimal implementation, so I haven't tested it that thoroughly. 
Maybe we could open a seperate issue to create evaluation scripts for this feature?
I'll add some documentation for the variables Monday.

vessemer:
@WGierke, I guess, the Hausdorff distance will suits it well. If you are mind to open an issue aimed at segmentation evaluation, I'll be glad to implement this feature.

prediction/src/algorithms/segment/src/levelset_segment.py
lamby: Mmm, please do so :)

prediction/src/algorithms/segment/src/levelset_segment.py
```diff
@@ -0,0 +1,46 @@
...
+def segment_nodule(image: sitk.Image, nodule: dict, lower_threshold=-600, upper_threshold=500):
+    """
+
+    :param image: Input image
+    :param lower_threshold: Lower Hu threshold of the nodule to be segmented. Should be about the Hu of lung
+    :param upper_threshold: Upper Hu threshold of the nodule to be segmented. Should be less than the Hu of bone
+    :param nodule: dict containing the nodule coordinates
+    :return:
+    """
+
```
reubano: Would you mind using [Google Style Docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)?

dchansen: Sorry. Oversight on my part. Fixed.

prediction/src/algorithms/segment/src/levelset_segment.py
```diff
@@ -0,0 +1,52 @@
...
+    seg = sitk.BinaryDilate(seg, 3)
+    init_ls = sitk.SignedMaurerDistanceMap(seg, insideIsPositive=True, useImageSpacing=True)
+
+    lsFilter = sitk.ThresholdSegmentationLevelSetImageFilter()
+    # Sets the Hu value range for the nodule
+    lsFilter.SetLowerThreshold(lower_threshold)
+    lsFilter.SetUpperThreshold(upper_threshold)
+    # Sets the error tolerance and number of iterations for the levelset solver
+    lsFilter.SetMaximumRMSError(0.02)
+    lsFilter.SetNumberOfIterations(1000)
```
lamby: I think someone previously mentioned here in review about explaining or making these constants/or at least documenting the magic numbers a bit? See also ``BinaryDilat(seg, 3)``. Three!

prediction/src/algorithms/segment/trained_model.py
```diff
@@ -36,15 +37,31 @@ def predict(dicom_path, centroids):
...
+
+    volumes = []
+    diameters = []
+    for nodule in centroids:
+        mask, diameter, volume = levelset_segment.segment_nodule(image, nodule)
+        volumes.append(volume)
+        diameters.append(diameter)
+        binary_mask |= sitk.GetArrayFromImage(mask).astype(bool)
+
     segment_path = os.path.join(os.path.dirname(__file__),
```
lamby: Out of interest, don't we have some kind of ``settings.TEST_DATA`` directory setting? :)

dchansen: Not to my knowledge? But this is not test data at all - this is the binary mask produced by the algorithm. Ideally, we would use a temporary file, but someone has to clean it up after use, so I thought it would be better to just reuse the same filename.

prediction/src/algorithms/segment/src/levelset_segment.py
dchansen: @lamby, I am a little unsure what you are asking.
The algorithm is iterative, and RMSError and number of iterations are the stopping criteria. None of the parameters are "special", and it seems a bit weird to have a line like
```
maximumRMSError = 0.02
lsFilter.SetMaximumRMSError(maximumRMSError)
```
Most of the parameters were taken from the SITK tutorial on segmentation, but some (curvature and thresholds) were changed to work better for lung segmentation. 
The parameters are either very easy to understand (the thresholds), or quite mathematically involved (curvature).  I've tried to add a few more comments, but as I said, I'm  not entirely sure what you're looking for.

reubano: In this case it would be upper case for constants.... so `MAX_RMS_ERROR` perhaps. It may look weird at first, but it's usually best to declare constants at the top of the function or file.

prediction/src/algorithms/segment/trained_model.py
reubano: What's wrong with a tempfile? Specifically, `tempfile.NamedTemporaryFile`

> This function operates exactly as TemporaryFile() does, except that the file is guaranteed to have a visible name in the file system (on Unix, the directory entry is not unlinked). That name can be retrieved from the name attribute of the returned file-like object. Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later). If delete is true (the default), the file is deleted as soon as it is closed.

prediction/src/algorithms/segment/src/levelset_segment.py
lamby: :+1:

prediction/src/algorithms/segment/trained_model.py
dchansen: The only problem with tempfile is who has the responsibility of removing it? The client would then have to do the cleanup, wouldn't it?
I can certainly implement this, but I just didn't want us ending up in a situation where we have a harddrive filled with random binary masks.

prediction/src/algorithms/segment/src/levelset_segment.py
dchansen: Hmm... not quite sure it increases readability in this case. After more consideration, I've moved the parameters to the args of the function, with reasonable defaults.

prediction/src/algorithms/segment/trained_model.py
dchansen: Well, more likely we would close the file after writing, which would delete it before the client could load it.

dchansen: Well, more likely we would close the file after writing, which would delete it before the client could load it.

lamby: I've implemented something similar here:

https://anonscm.debian.org/git/reproducible/diffoscope.git/tree/diffoscope/tempfiles.py#n29

.. notice how it adds files to a global which is called when we exit.

dchansen: Ok, I've done something similar, and added cleanup via atexit. Still means we will have to restart the service occasionally, but I guess that shouldn't be a problem.

prediction/src/algorithms/segment/trained_model.py
```diff
@@ -36,20 +38,45 @@ def predict(dicom_path, centroids):
...
+    volumes = []
+    diameters = []
+    for nodule in centroids:
+        mask, diameter, volume = levelset_segment.segment_nodule(image, nodule)
+        volumes.append(volume)
+        diameters.append(diameter)
+        binary_mask |= sitk.GetArrayFromImage(mask).astype(bool)
+
+    segment_path = tempfile.NamedTemporaryFile(suffix='.npy')
+    _FILES.append(segment_path)
```
lamby: Sorry to mess you around here, but once I saw this implementation I just can't quite help feeling it doesn't feel right. :(

I'm not really sure about the best solution. Can't we just pass around the ``TemporaryFile`` object? Or change the API such that people are passing it *into* this method? At the very very least we should factor our the ``_FILES`` mangling into a global utility (eg. ``get_named_temporary_file``) that isn't local to this module.

Other suggestions highly highly welcome!

dchansen: The problem is that our output is just the path to the mask - we have no way of knowing when the client is done with the file. 
I think it makes sense to move this into a global utility instead.

reubano: Just to throw out another idea, how about having a `segment_path` kwarg? something like:

```python
def predict(*args, segment_path='default_path.npy'):
    pass
```

dchansen: @lamby Where would we want such a util to live? prediction/src/utils.py?

@reubano I think we still have the fundamental problem that either the client needs to clean up, or the prediction service does?

lamby: @dchansen Sounds reasonable, does that "feel" good when you look at the code? Could probably live one level up if we have shared utils...?

dchansen: Ok. Moved the functionality into utils. I think this is a reasonable solution?

reubano: Looks good... but you still need the exit handler right? FYI, it also works as a decorator. From the docs:

```python
import atexit

@atexit.register
def goodbye():
    print("You are now leaving the Python sector.")
```

dchansen: No, that was a mistake on my part. All the files are closed (and thus deleted) when the program exits, so this should be ok.

lamby: > No, that was a mistake on my part. All the files are closed (and thus deleted) when the program exits, so this should be ok.

Can you add that to the docstring to the util you created?

Also, what's the point of `_FILES` if we aren't cleaning up?

lamby:
> No, that was a mistake on my part. All the files are closed (and thus deleted) when the program exits, so this should be ok.

Can you add that to the docstring to the util you created? Wait, what's the point of _FILES if we aren't cleaning up then?

I'm confused :)

lamby:
@dchansen Gentle ping here :)

dchansen:
I am closing the pull request. 
The point of it was to implement something simple, to quickly get the prediction service fully featured - 19 days later I think we somewhat missed that.  
It is also directly competing with pull request [#147](https://github.com/concept-to-clinic/concept-to-clinic/pull/147), but with no objective way to compare them (we can't use the LIDC data).
Either way, whichever pull request does not get selected will essentially be wasted work.

reubano:
Hi @dchansen please see [my comment here](https://github.com/concept-to-clinic/concept-to-clinic/pull/147#issuecomment-338402309).

Issue #167, opened 2017-10-16
Use css preprocessor like sass, less...


truefalse10:
The current vue.js setup allows only to write styles in pure css. This can be quite time consuming, error-prone and leads to redundant code. In order to use variables, mixins and all other handy tools I would suggest using a preprocessor. In vuejs this can be easily achieved by installing the preferred webpack [loader](https://vue-loader.vuejs.org/en/configurations/pre-processors.html). My suggestions would be to use SCSS because it has the same syntax as CSS and can therefore easily be adapted by people new to preprocessores. 

## Expected Behavior
I can write component styles with scss/sass/less/stylus and it gets compiled to css

## Current Behavior
using `<style lang="scss">` leads to compile time error

## Possible Solution
install missing webpack loaders

WGierke:
So has this now been implemented in 11d04b0?

calvernaz:
Exactly what is missing here?

isms:
@lamby Can you look into this and close the issue if necessary?

PR #168, opened 2017-10-16
Install sass-loader, refactor .vue components to use scss #167

truefalse10:
Will install sass-loader and update all components to use scss. #167 

# Motivation
The current vue.js setup allows only to write styles in pure css. This can be quite time consuming, error-prone and leads to redundant code. In order to use variables, mixins and all other handy tools I would suggest using a preprocessor. In vuejs this can be easily achieved by installing the preferred webpack [loader](https://vue-loader.vuejs.org/en/configurations/pre-processors.html). My suggestions would be to use SCSS because it has the same syntax as CSS and can therefore easily be adapted by people new to preprocessores. 

## Expected Behavior
I can write component styles with scss/sass/less/stylus and it gets compiled to css

## Current Behavior
using `<style lang="scss">` leads to compile time error

## Possible Solution
install missing webpack loaders


lamby:
Thanks!

Issue #162, opened 2017-10-16
Fix UnhandledPromiseRejectionWarning in Vue frontend 


reubano:
## Overview
Currently, when running `docker-compose -f local.yml up vue`, the following error appears:

```bash
vue_1         |  DONE  Compiled successfully in 12582ms20:03:55
vue_1         | 
vue_1         | > Listening at http://0.0.0.0:8080
vue_1         | 
vue_1         | (node:15) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Exited with code 3
vue_1         | (node:15) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

## Expected Behavior
The error should not occur.

## Technical details

- This feature should be implemented in the [`interface/frontend/src/`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/interface/frontend/src) folder.

## Acceptance criteria

 - [ ] running `docker-compose -f local.yml up vue` should exit with no errors. 

*NOTE: All PRs must follow the standard PR checklist.*

lamby:
Fixed! Thanks!

PR #171, opened 2017-10-17
Fixing error unhandled promise rejection #162, removing VueResource

truefalse10:
as described in #162 with node 8 an error is thrown after starting the vuejs dev-server. This is caused by the dev-server trying to start a browser window inside the docker container which leads to a rejected promise.

This PR fixes this issues. Additionally it improves the following:
* remove VueResource (we use axios for http requests as i understood, therefore VueResource is not needed anymore)
* fix current axios requests to also handle rejected promises correctly with catch statements
* add correct node_modules path to .dockerignore

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,5 @@
 .*
 !.env
 !setup.cfg
-node_modules
+interface/frontend/node_modules
 npm-debug.log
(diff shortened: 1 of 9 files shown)


lamby:
Merged, many thanks!

Issue #149, opened 2017-10-05
Vue: implement nodule 'Annotation' UI component(s)


isms:
### This issue will remain open until all the checkboxes are checked! PRs can reference this issue and will get credit even if the whole issue is not yet closed :+1:

Note: split off from #110 

---

Design doc reference here: http://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#annotate

Mock-up:
![](http://concept-to-clinic.readthedocs.io/en/latest/_images/wireframe-2-1-annotation-and-segmentation-classify.svg)

## Expected Behavior
Users should be able to annotate nodules marked concerning and classify them as concerning or not concerning.

- [ ] Provide UI controls for adding sending general RSNA template notes for the case.
- [ ] Provide UI controls for selecting a nodule for detail view for a single nodule and display that nodule in a corresponding image viewer.
- [ ] In the detail view for a single nodule, provide a UI control for specific a user-supplied ground truth label for P(concerning). Also show an indication of what the predicted P(concerning) was for each nodule.
- [ ] In the detail view, provide UI controls for the nodule specific RSNA template data outlined above.
- [ ] In the detail view, enrich the image viewer with pan, zoom, and slide traversal for taking a closer look at and around nodules.

## Mocking data

Feel free to mock data for `state` if it is not yet implemented in the backend API, and feel free to `POST` to API endpoints that don't exist yet :sparkles:

louisgv:
I will start working on this component, a WIP PR will be issued soon.

louisgv:
@isms How can I obtain need a mock sample json for a nodule?

Or the better question would be the documentation for the json schema that would be used to describe a nodule

The `/api/nodules.json` endpoint is also returning an empty array.

![image](https://user-images.githubusercontent.com/6723574/31361807-307a217a-ad23-11e7-8e83-6e982567ed7f.png)

louisgv:
Also, it seems to me that there is already a NoduleList component in charge of rendering the list of Nodule to be examine, it is however seems to be lacking most of the UI desired in the mock. 

Should I implement those feature as a separate component and attach them to the Nodule.vue component, or should I refactor the Nodule list to render instead an `Annotate` component?

I have

isms:
@louisgv Here's how I can get some data fixtures:

First use the testing fixture factories to create some data in your local database:
```
docker-compose -f local.yml run interface python manage.py shell_plus

[...]

PR #173, opened 2017-10-18
Initialize 'Annotation' UI component

louisgv:
Per #149, I implemented 3 components for the annotation: The Annotation editor itself, a ConcernSlider component and a RadioButton component, all are generic and reusable. I also altered the behavior of `Nodule` component so that it is more flexible, allowing foreign editor (like the Annotation editor) to be inserted in the form of a `slot`

## Description

+ Implement Annotation component
+ Implement ConcernSlider component (generic)
+ Implement RadioButton component (generic)
+ Refactor Nodule component

#149 

## Motivation and Context

Design doc reference here: http://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#annotate

## How Has This Been Tested?

+ It seems front-end is lacking test standard/example so I will refrain from making unnecessary architectural decision (unless it is called for).
+ I tested this briefly with other component and it works ok.
+ This will heavily alter the behavior of the `Nodule` component. 

image

image

## CLA
- [+ ] I have signed the CLA; if other committers are in the commit history, they have signed the CLA as well


interface/frontend/src/assets/css/project.css
```diff
@@ -4,4 +4,16 @@ div#version {
   top: .5rem;
   right: 1rem;
   z-index: 10000;
-}
\ No newline at end of file
+}
+
+div.container {
```
truefalse10: Do we really want this global style for all div.container? the class container might be used by other developers and would then produce some unexpected behavior

lamby:
> Beside the global styles in project.css the PR looks good to me.

Same here :)

louisgv:
@truefalse10 @lamby  thanks for the feedback. I re-scoped those style so that it only affect AnnotateAndSegment. 

However I would highly recommend a global style for this so that future development will not have to worry about the nav bar. Indeed it shouldn't be using the className `container`, as that will potentially conflict with bootstrap. Regardless, I think that all `view` should have an uniformed offset from the nav.

lamby:
Thanks!

Issue #157, opened 2017-10-15
Give interface backend a home route


reubano:
## Overview
Currently, navigating to `http://0.0.0.0:8000/` returns a `404` error.

![screen shot 2017-10-16 at 1 33 29 am](https://user-images.githubusercontent.com/157864/31589968-14d235c2-b212-11e7-9553-8c79db8fa82e.png)

## Steps to Reproduce

1. `docker-compose -f local.yml up interface`
2. `open http://0.0.0.0:8000/`

## Expected Behavior

This url should route to a page listing all the available routes

## Technical details

- This feature should be implemented in the [`interface/backend/api/`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/interface/backend/api) folder.

## Acceptance criteria

 - [ ] Navigating to `http://0.0.0.0:8000/` should produce a helpful page with links to the available routes.

*NOTE: All PRs must follow the standard PR checklist.*

vessemer:
Is there something else left that prevents this issue from getting closed?

reubano:
nope... thanks for the reminder

PR #174, opened 2017-10-19
#157 Listed available routes

vessemer:
This `http://0.0.0.0:8000/` route to a page listing all the available routes.

## Description
`http://0.0.0.0:8000/` redirect to `/api/routes` which in turn will return the dictionary of available links. 

This addresses #157.

## Motivation and Context
Currently, navigating to `http://0.0.0.0:8000/` returns a 404 error.

## Screenshots:
The API and JSON response:
screenshot from 2017-10-19 16-46-23
screenshot from 2017-10-19 16-47-42

## CLA

### DIFF
--- a/interface/config/urls.py
+++ b/interface/config/urls.py
@@ -4,9 +4,11 @@ from django.conf.urls import (
     url
 )
 from django.views.static import serve
+from django.views.generic import RedirectView
 
 urlpatterns = (
     url(r'^api/', include('backend.api.urls', namespace='')),
+    url(r'^', RedirectView.as_view(url='api/'), name="available-routes"),
 )
 
 if settings.DEBUG:


isms:
How about just a redirect from `/` to `/api` since that shows all the available endpoints? Scraping the source code for url patterns doesn't seem like a great long term answer.

vessemer:
@isms, I have to agree with you, it seems that it fits the requirements of the issue #157. 
@reubano can you take a look, please?

vessemer:
@isms, done: all squashed and rebased.

lamby:
Merging, but I think I would have liked to have seen the "urlconf" instead of the raw ``/api`` in the redirect ;)

Issue #158, opened 2017-10-15
Remove hidden files from Vue frontend


reubano:
## Overview
Currently, when you navigate to `http://0.0.0.0:8080/` and click to expand the folders, you see hidden files such as `. DS_Store`.

![screen shot 2017-10-14 at 11 02 12 pm](https://user-images.githubusercontent.com/157864/31590120-c04effa6-b213-11e7-9ce7-50b7aa5a4d3c.png)

## Steps to Reproduce

1. `docker-compose -f local.yml up vue`
1. `open http://0.0.0.0:8080/`
1. Expand the pluses to navigate into the folders

## Expected Behavior

The hidden files should not be visible

## Technical details

- This feature should be implemented in the [`interface/frontend/src/`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/interface/frontend/src) folder.

## Acceptance criteria

 - [ ] Navigating to `http://0.0.0.0:8080/` and expanding the folders should not reveal any hidden files.

*NOTE: All PRs must follow the standard PR checklist.*

jjjmm:
wouldn't it be better to filter directories in the interface backend, before sending them to the frontend?

reubano:
Good idea @jjjmm! I've added the backend label as well now.

jjjmm:
@reubano btw, is there any instructions, how can i debug interface-backend in docker?

reubano:
@jjjmm try `docker-compose -f local.yml up --build interface`

lamby:
Merged :)

PR #175, opened 2017-10-21
#158 Hide hidden files

vessemer:
Filtering out the hidden files from the Vue frontend.

During working over this PR I've stumbled upon the unexpectable behaviour in `src/components/ImageSeries.vue`:
```javascript
[Vue warn]: Error in render: "TypeError: _vm.availableSeries is undefined"

found in

---> <ImageSeries> at src/components/ImageSeries.vue
       <OpenImage> at src/views/OpenImage.vue
         <App> at src/App.vue
           <Root>
```
Though the response of the `/api/images/available` was sustainable and in a proper format. It took a while from me to figure out that this behaviour was introduced by the PR #171, specifically by the exchange `$http` on `$axios` ([1](https://github.com/concept-to-clinic/concept-to-clinic/pull/171/files?diff=unified#diff-ff7a640b74e49c1a5b1dbb329689f4b7R102) and [2](https://github.com/concept-to-clinic/concept-to-clinic/pull/171/files?diff=unified#diff-ff7a640b74e49c1a5b1dbb329689f4b7R115) diffs). Thus I roll back these changes, of course, with no disruption from `UnhandledPromiseRejection`. The output of `docker-compose -f local.yml up vue` does not contain errors: 
```bash
vue_1            | [HPM] Proxy created: /  ->  http://interface:8000
vue_1            | > Starting dev server...
vue_1            |  DONE  Compiled successfully in 3260ms02:15:12
vue_1            | 
vue_1            | > Listening at http://0.0.0.0:8080
vue_1            | 
[]
```
## Description
The filtering is done on the backend side. Since there is nothing like `os.ishidden` I've decided to not over complicate and wrote the static function [`ImageAvailableApiView.is_hidden`](https://github.com/concept-to-clinic/concept-to-clinic/compare/master...vessemer:158_hide_hidden_files?expand=1#diff-500c04290e5206134d43cf7c48c3976bR80).

This addresses the issue #158.

## Comparison:

screenshot from 2017-10-21 04-00-23
screenshot from 2017-10-21 03-47-12

## CLA

### DIFF
--- a/interface/backend/api/views.py
+++ b/interface/backend/api/views.py
@@ -1,6 +1,7 @@
 import json
 import mimetypes
 import os
+
 import dicom
 from backend.api import serializers
 from backend.cases.models import (
@@ -13,8 +14,8 @@ from backend.images.models import ImageSeries
 from django.conf import settings
 from django.core.files.storage import FileSystemStorage
 from django.shortcuts import get_object_or_404
-from rest_framework.decorators import api_view
 from rest_framework import viewsets
+from rest_framework.decorators import api_view
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
@@ -40,7 +41,6 @@ class ImageSeriesViewSet(viewsets.ModelViewSet):
 
 
 class ImageMetadataApiView(APIView):
-
     def get(self, request):
         '''
         Get metadata of a DICOM image including the image in base64 format.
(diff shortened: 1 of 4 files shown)


interface/backend/api/views.py
```diff
@@ -76,15 +76,28 @@ def filename_to_dict(name, location):
...
-        tree['children'] = [self.walk(os.path.join(location, dir), dir) for dir in folders]
+        tree = None
+        if not self.is_hidden(location):
+            folders, files = self.fss.listdir(location)
+            tree = {'name': dir_name, 'children': []}
+            tree['files'] = [self.filename_to_dict(filename, location)
+                             for filename in sorted(files)
+                             if not self.is_hidden(filename)]
+            tree['type'] = 'folder'
+            tree['children'] = [self.walk(os.path.join(location, dir), dir)
```
lamby: I think this needs a ``sorted``

interface/backend/api/views.py
```diff
@@ -76,15 +76,28 @@ def filename_to_dict(name, location):
...
         """
-        folders, files = self.fss.listdir(location)
-        tree = {'name': dir_name, 'children': []}
-        tree['files'] = [self.filename_to_dict(filename, location) for filename in sorted(files)]
-        tree['type'] = 'folder'
-        tree['children'] = [self.walk(os.path.join(location, dir), dir) for dir in folders]
+        tree = None
+        if not self.is_hidden(location):
+            folders, files = self.fss.listdir(location)
+            tree = {'name': dir_name, 'children': []}
```
lamby: Why not just define ``tree = { .. }`` all in one go, like:

```
tree = {
    'name': dir_name
    'type': 'folder',
    'children': […],
    'files': […],
}
```

Assigning and then re-assigning makes me think (incorrectly) that we are *re-using* values which is misleading

vessemer: `tree` is now defined in one go, additionally, I've assumed that initial location shall not be hidden.

interface/backend/api/views.py
vessemer: Corrected, thanks!

Issue #180, opened 2017-10-23
CSS/HTML template: adjust `container` margins to give more screen real estate


isms:
We want some more room to work with:
![image](https://user-images.githubusercontent.com/1284973/31911268-ee12fe24-b7f4-11e7-9f34-f917c213b370.png)

lamby:
Merged, thank you!

PR #184, opened 2017-10-23
#180 CSS adjust margin

vessemer:
> We want some more room to work with

This address #180 

## Screenshots:
screenshot from 2017-10-24 00-10-29
screenshot from 2017-10-24 00-09-50

## CLA


vessemer:
@lamby, only now I noted that this issue was `FIRST PR ONLY`, due to the lack of sleep treat it as `POINT BOUNTY`, my excuse.

Issue #181, opened 2017-10-23
Add fields to `Case` and `Nodule` models based on RadReport template


isms:
RadReport defines industry best practices for what information should be recorded on a lung cancer screen. Some are specific to the overall image series (the `Case` model) and some are specific to each nodule (the `Nodule` model).

## Expected Behavior
Appropriate new fields on appropriate models with the migrations that implement them.

## References
[1] Design doc - http://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#results
[1] Rad Report template for CT Lung Cancer Screening - http://www.radreport.org/template/0000268

lamby:
Merged in #185.. thanks!

PR #185, opened 2017-10-24
#181 RadReport's template fields

vessemer:
The models were adopted to suit the [RSNA Radiology Reporting Templates](http://www.radreport.org/template/0000268) for CT Lung Cancer Screening.

## Description
All feilds were assigned to the [`Case`](https://github.com/concept-to-clinic/concept-to-clinic/compare/master...vessemer:181_RadReport_template_fields?expand=1#diff-421ccf31e89cb76aa475b257a8904f4eR55) model, except for those listed below which were assigned to the [`Nodule`](https://github.com/concept-to-clinic/concept-to-clinic/compare/master...vessemer:181_RadReport_template_fields?expand=1#diff-421ccf31e89cb76aa475b257a8904f4eL41) model: 
- `diameter`
- `appearance_feature`
- `density_feature`
- `image_no`  

Due to the symmetry of right and left pleural spaces, a new model [`PleuralSpace`](https://github.com/concept-to-clinic/concept-to-clinic/compare/master...vessemer:181_RadReport_template_fields?expand=1#diff-421ccf31e89cb76aa475b257a8904f4eR19) was created and it consists of:
- `effusion`
- `calcification`
- `thickening`
- `pneumothorax`  

The [`Case`](https://github.com/concept-to-clinic/concept-to-clinic/compare/master...vessemer:181_RadReport_template_fields?expand=1#diff-421ccf31e89cb76aa475b257a8904f4eR55)  model then receives two foreign keys for left and right pleural spaces.

This address #181 

## CLA


interface/backend/cases/models.py
```diff
@@ -10,6 +10,48 @@
...
+    cls.do_not_call_in_templates = True
+    return cls
+
+
+class PleuralSpace(models.Model):
+    created = models.DateTimeField(default=timezone.now)
+
+    @unique  # ensures all variables are unique
+    @django_enum
+    class TypesOne(IntEnum):
```
lamby: From experience I have a very strong preference for putting enumerations into a separate ``enums.py`` file alongside the ``models.py`` — could you try this approach and see how it looks/works for you? It might not be suitable, but definitely worth having a look at...

interface/backend/cases/models.py
```diff
@@ -10,6 +10,48 @@
...
+    @unique  # ensures all variables are unique
+    @django_enum
+    class TypesTwo(IntEnum):
+        NONE = 0
+        MILD = 1
+        MODERATE = 2
+        EXTENSIVE = 3
+
+    effusion = models.IntegerField(
+        choices=[(choice.value, choice.name.replace("_", " ")) for choice in TypesOne],
```
lamby: Feels like we could abstact this :)

interface/backend/cases/models.py
```diff
@@ -18,6 +60,104 @@ class Case(models.Model):
...
+                                            related_name="right_pleural_space",
+                                            null=True)
+
+    heart_size = models.IntegerField(
+        choices=[(choice.value, choice.name.replace("_", " ")) for choice in HeartShapeChoices],
+        default=HeartShapeChoices.NORMAL.value)
+
+    coronary_calcification = models.IntegerField(
+        choices=[(choice.value, choice.name.replace("_", " ")) for choice in ShapeChoices],
+        default=ShapeChoices.NONE.value)
```
lamby: I worry that we have too many nullable (or quasi-nullable with ``NONE`` enum value) here. These are pretty common sources of bugs. Can I suggest trying a "hanging table" approach and seeing how that feels? eg.

```
class Parent(models.Model):
    field_a = models.IntegerField()
    field_b = models.IntegerField()

class ExtraDataForParent(models.Model):
    parent = models.OneToOneField(Parent, related_name='extra_data', primary_key=True)
    field_c = models.IntegerField()
    field_d = models.IntegerField()
```
Note that crucially  ``field_c`` and ``field_d`` are not nullable as you only create ``ExtraData`` models when you know you have both of those fields, so there is no possibility of having ``field_c`` but not ``field_d`` etc. etfc.

vessemer: Thank you, it's absolutely rational to decompose the models.

interface/backend/cases/models.py
vessemer: Done, I've added the [`format_enum`](https://github.com/concept-to-clinic/concept-to-clinic/pull/185/commits/d4e5a767b0dc943a624d9ee577a5a1dbe990ac4d#diff-11ad5b67ec28e05935f1d6da87a683e8R10) function inside the [`enum.py`](https://github.com/concept-to-clinic/concept-to-clinic/pull/185/commits/d4e5a767b0dc943a624d9ee577a5a1dbe990ac4d#diff-11ad5b67ec28e05935f1d6da87a683e8) file.

interface/backend/cases/models.py
vessemer: Done: [`enum.py`](https://github.com/concept-to-clinic/concept-to-clinic/pull/185/commits/d4e5a767b0dc943a624d9ee577a5a1dbe990ac4d#diff-11ad5b67ec28e05935f1d6da87a683e8), looks much better now, thanks!

lamby:
Nice @vessemer. What do you think of the separate enum file approach? The only thing I would ask now is that you rename it to ``enums.py`` as Python already ships with an ``enum`` module so this clashes... \*g*

lamby:
Merged, many thanks!

isms:
Awesome - thanks @vessemer :tada:

Issue #179, opened 2017-10-23
Vue: reusable Javascript DICOM image viewer component


isms:
Splitting this off from #148 and #150.

### **Note:** Proofs of concept with a specific library will be in a position to receive points for this even if a different way of doing it is ultimately selected. These add value to the project by giving other contributors a reference point for how this may be done.

We need to be able to show DICOM imagery in an interactive way. There are existing libraries designed to do this:
* [Cornerstone Tools](https://github.com/chafey/cornerstoneTools) (see especially [this example](https://rawgit.com/chafey/cornerstoneTools/master/examples/freehandRoi/index.html) of drawing freehand boundaries)
* ~DICOM web viewer~ (GNU GPL v3 not compatible with MIT license)
* Others (Google shows a few more)

## Expected Behavior

Somebody needs to make these work nicely with Vue so we can show things like predicted nodule centroids (#148) and boundaries (#150) above the image slices, and also let them change centroids or mark new places. See the design doc for a discussion of features.

isms:
@swederik Any interest in taking a crack at this?

swederik:
I don't have the time at the moment, sorry.

jjjmm:
@isms i could give it a try, since it is within scope of #150

jjjmm:
@isms [DICOM web viewer](https://ivmartel.github.io/dwv/) is under GNU GPL 3 license. Does it mean, that we can't use it in this project, since it is under MIT license?

isms:
@jjjmm Good point: GNU GPL 3 not compatible with MIT. I'll scratch that.

Issue #150, opened 2017-10-05
Vue: implement 'Segmentation' UI component(s)


isms:
### This issue will remain open until all the checkboxes are checked! PRs can reference this issue and will get credit even if the whole issue is not yet closed :+1:

Note: split off from #110 

---

Design doc reference here: http://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#segment

Mock-up:
![](http://concept-to-clinic.readthedocs.io/en/latest/_images/wireframe-2-2-annotation-and-segmentation-segment.svg)

## Expected Behavior

- [ ] Display an image view of a given slice of the selected nodule. Allow navigation between slices.
- [ ] Overlay the current vertices on top of the nodule image to see the predicted boundary.
- [ ] Allow the user to move existing vertices.
- [ ] Allow the user to delete existing vertices.
- [ ] Allow the user to add a new vertex splitting a line segment connecting two vertices.
- [ ] Allow the user to extend the nodule boundary along the Z axis by extending the currently terminal polygon up or down to the adjacent slice which still has unhighlighted nodule.
- [ ] Indicate to the user the Z axis extent of slices.

## Mocking data

Feel free to mock data for `state` if it is not yet implemented in the backend API, and feel free to `POST` to API endpoints that don't exist yet :sparkles:

swederik:
In Cornerstone Tools (https://github.com/chafey/cornerstoneTools) we have an ROI tool that could be a good starting point for building the nodule boundary selection tool: 
https://rawgit.com/chafey/cornerstoneTools/master/examples/freehandRoi/index.html

jjjmm:
i am planning to start implementing this component on this weekend, if no one wants to start earlier

jjjmm:
@isms I have a question about Segmentation component. Should it be in the sync with Annotate component? When nodule in Annotate selected, should it be also selected in Segmentation component? If so, could i put both Annotate and Segmentation components under the same Nodule (as on added picture)? If i understood correctly, both components represent data of same nodule ![segmentation](https://user-images.githubusercontent.com/17836809/31794931-96876f48-b52c-11e7-8a20-0180e86aacac.png)
 Or selection actions should be independent and implemented as on current design mock?

isms:
@jjjmm Yes, they should be in sync, the segmentation screen on the right is showing the same nodule so that the radiologist can see what they are annotating and validate or redraw the predicted borders.

Thanks for asking!

[...]

Issue #148, opened 2017-10-05
Vue: implement 'Detect and Select' component(s)


isms:
### This issue will remain open until all the checkboxes are checked! PRs can reference this issue and will get credit even if the whole issue is not yet closed :+1:

Note: split off from #110, interaction with imagery tracked in #179 

---

Design doc reference here: http://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#detect-and-select

Mock-up:
![](http://concept-to-clinic.readthedocs.io/en/latest/_images/wireframe-1-detect-and-select.svg)

## Expected Behavior
Users should be able to view predictions of nodule centroids and mark whether they are concerning or not:

- [ ] List all available predicted candidate sites. Allow each to be selected to view details.
- [ ] When a candidate is selected, show a detail view in an image viewer control. By default, show the slice containing the predicted centroid and a marker clearly indicating where the predicted centroid occurs.
- [ ] Allow the user to zoom, pan, and navigate through slices in the image viewer.
- [ ] Allow the user to window grayscale levels.
- [ ] Allow the user to toggle the candidate centroid marker’s visibility.
- [ ] Allow the user to mark any predicted candidate for further analysis.
- [ ] Allow the user to navigate through the imagery freestyle and mark other locations which the models missed.
- [ ] When the user has finished marking candidates, send all of this labeled data to the backend in the format specified above.

## Possible Implementation
Basic Django templates were created for this in #88 - that is probably a good place to start.

## Mocking data

Feel free to mock data for `state` if it is not yet implemented in the backend API, and feel free to `POST` to API endpoints that don't exist yet :sparkles:

lmiller1990:
Hi. I love what the project is looking to achieve and am a great supporter of the cause. I've done a lot of Vue.js, so I'm going to try and start building some of the UI related issues.

isms:
@lmiller Glad to have you on board! :boat:

lmiller1990:
Thanks. I'll see if I can get something done in the next few days, got everything up and running plus created some fake data.

Project is still early days - there aren't any models or much else in terms of data set up yet, should I just fake/make it up for now?

stranger26:
@lmiller1990 I think you should just fake it, at least that is from what I understand from the 'Mocking data' section.

isms:
@lmiller1990 @stranger26 New info on getting test data [here] - I also opened #153 to publicize this more.

swederik:
[...]

PR #188, opened 2017-10-26
#179 DICOM image viewer

vessemer:
The ability to show DICOM images is one of the most important parts for this project. There are existing libraries that provide a way to show DICOM images that can help with this task. It were two main candidates with proper library to choose from: [ivmartel](https://github.com/ivmartel) and [chafey](https://github.com/chafey). The decision was made to use the [cornerstone](https://github.com/chafey/cornerstone) library for the reason that it flexible and can provide the project with ability to show images in an interactive way. It is also was much easier to use with VueJS than its competitor.

Two main difficulties were met during the work on this issue. The first one was the problem with pixelData serialization. It took some time before a good solution was found.

The second problem was to make library work correctly with VueJS. Choosing the most appropriate llibrary also refers to this problem. A lot of time was spent on testing and studying different libraries and browsing repositories before the appropriate solution was found.

## Description

This is the implementation of DICOM image viewer that was made with the help of [cornerstone](https://github.com/chafey/cornerstone) and [cornerstoneTools](https://github.com/chafey/cornerstoneTools). The image is being taken by the part of the url that adresses it.

This address #179

## Motivation and Context
There is a strong need in the way of showing DICOM imagery. It is impossible to make that project at least somehow usable by common people without DICOMs actually being shown. Furthermore, issues [#148](https://github.com/concept-to-clinic/concept-to-clinic/issues/148) and [#150](https://github.com/concept-to-clinic/concept-to-clinic/issues/150) depend on it.

## CLA

### DIFF
--- a/interface/frontend/src/components/open-image/OpenDICOM.vue
+++ b/interface/frontend/src/components/open-image/OpenDICOM.vue
@@ -1,0 +1,132 @@
+<template>
+  <div class="container">
+    <div class="col-xs-6">
+      <div class="DICOM-container">
+        <div class="DICOM" ref="DICOM"></div>
+      </div>
+      <p>{{ dicom.imageId}}</p>>
+    </div>
+  </div>
+</template>
+
+<script>
+  var cornerstone = require('cornerstone-core')
+  var Q = require('q')
+
+  export default {
+    name: 'open-dicom',
+    data () {
+      return {
+        dicom: {
+          imageId: '',
+          minPixelValue: 0,
+          maxPixelValue: 255,
+          slope: 1.0,
+          intercept: 0,
+          windowCenter: 90,
+          windowWidth: 100,
+          render: cornerstone.renderGrayscaleImage,
+          getPixelData: this.getPixelData,
+          rows: 512,
+          columns: 512,
+          height: 512,
+          width: 512,
+          color: false,
+          base64data: '',
+          columnPixelSpacing: 1,
+          rowPixelSpacing: 1,
+          sizeInBytes: 512 * 512 * 2
+        },
+        dicomUrl: 'LIDC://LIDC-IDRI-0001/1.3.6.1.4.1.14519.5.2.1.6279.6001.298806137288633453246975630178/1.3.6.1.4.1.14519.5.2.1.6279.6001.179049373636438705059720603192/-95.000000.dcm',
+        info: null,
+        csName: 'LIDC'
+      }
+    },
+    watch: {
+      info: function (val) {
+        if (val != null) {
+          this.applyMeta(val)
+          console.log(this.dicom)
+          this.loadImage(this.resolveDICOM)
+        }
+      }
+    },
+    mounted: function () {
+      this.fetchData(this.dicomUrl)
+    },
+    methods: {
+      fetchData (id) {
+        this.$axios.get('/api/images/metadata?dicom_location=/images/' + id.slice(this.csName.length + 3))
+          .then((response) => {
+            this.info = response.data
+          })
+          .catch(() => {
+            // TODO: handle error
+          })
+      },
+      applyMeta (info) {
+        this.dicom.imageId = this.dicomUrl
+        var meta = info['metadata']
+        this.dicom.base64data = info['image']
+        this.dicom.slope = meta['Rescale Slope']
+        this.dicom.rows = meta['Rows']
+        this.dicom.columns = meta['Columns']
+        this.dicom.height = meta['Rows']
+        this.dicom.width = meta['Columns']
+        this.dicom.columnPixelSpacing = meta['Pixel Spacing']['0']
+        this.dicom.rowPixelSpacing = meta['Pixel Spacing']['1']
+      },
+      str2pixelData (str) {
+        var buf = new ArrayBuffer(str.length * 2) // 2 bytes for each char
+        var bufView = new Int16Array(buf)
+        var index = 0
+        for (var i = 0, strLen = str.length; i < strLen; i += 2) {
+          var lower = str.charCodeAt(i)
+          var upper = str.charCodeAt(i + 1)
+          bufView[index] = lower + (upper << 8)
+          index++
+        }
+        return bufView
+      },
+      getPixelData () {
+        var pixelDataAsString = window.atob(this.dicom.base64data)
+        var pixelData = this.str2pixelData(pixelDataAsString)
+        return pixelData
+      },
+      resolveDICOM (imageId) {
+        var deferred = Q.defer()
+        deferred.resolve(this.dicom)
+        return deferred.promise
+      },
+      loadImage (resolve) {
+        cornerstone.registerImageLoader('LIDC', resolve)
+        var element = this.$refs.DICOM
+        console.log(element)
+
+        cornerstone.enable(element)
+        console.log(resolve())
+        cornerstone.loadImage(this.dicom.imageId).then(function (image) {
+          cornerstone.displayImage(element, image)
+        })
+      }
+    }
+  }
+</script>
+
+<style lang="scss" scoped>
+  .DICOM-container {
+    width:512px;
+    height:512px;
+    position:relative;
+    display:inline-block;
+    color:white;
+  }
+
+  .DICOM {
+    width:512px;
+    height:512px;
+    top:0px;
+    left:0px;
+    position:absolute;
+  }
+</style>
--- a/interface/frontend/src/components/open-image/OpenDICOM.vue
+++ b/interface/frontend/src/components/open-image/OpenDICOM.vue
@@ -1,83 +1,89 @@
 <template>
   <div class="DICOM-container">
+    <div class="DICOM-description">{{ display }}</div>
     <div class="DICOM" ref="DICOM"></div>
   </div>
 </template>
 
 <script>
-  import { EventBus } from '../../main.js'
   const cornerstone = require('cornerstone-core')
   const Q = require('q')
 
   export default {
     name: 'open-dicom',
-    data () {
-      return {
-        dicom: {
-          imageId: '',
-          minPixelValue: 0,
-          maxPixelValue: 255,
-          slope: 1.0,
-          intercept: 0,
-          windowCenter: 110,
-          windowWidth: 100,
-          render: cornerstone.renderGrayscaleImage,
-          getPixelData: this.getPixelData,
-          rows: 512,
-          columns: 512,
-          height: 512,
-          width: 512,
-          color: false,
-          base64data: '',
-          columnPixelSpacing: 1,
-          rowPixelSpacing: 1,
-          sizeInBytes: 0
-        },
-        info: null,
-        csName: 'LIDC',
-        prefixUrl: '/api/images/metadata?dicom_location=/'
-      }
-    },
-    computed: {
-      'dicom.sizeInBytes': function () {
-        this.dicom.sizeInBytes = this.dicom.rows * this.dicom.columns * 2
-      }
-    },
-    watch: {
-      info: function (val) {
-        if (val != null) {
-          this.applyMeta(val)
-          this.loadImage(this.resolveDICOM)
+    props: {
+      view: {
+        type: Object,
+        default: {
+          type: 'DICOM',
+          prefixCS: ':/',
+          prefixUrl: '',
+          path: ''
         }
       }
     },
-    mounted: function () {
-      EventBus.$on('dicom-selection', (path) => {
-        this.dicom.imageId = this.csName + ':/' + path
-        console.log(this.dicom.imageId)
-        this.fetchData(this.dicom.imageId)
-      })
+    data () {
+      return {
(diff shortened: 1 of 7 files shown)


interface/frontend/src/views/OpenImage.vue
```diff
@@ -1,13 +1,129 @@
...
+        <div ref="DICOM" style="width:512px;height:512px;top:0px;left:0px; position:absolute;"></div>
+      </div>
+      <p>{{ dicom.imageId}}</p>>
+    </div>
+  </div>
 </template>
 
 <script>
 import ImageSeries from '../components/ImageSeries'
+var cornerstone = require('cornerstone-core')
```
louisgv: I suggest avoiding `var`.

interface/frontend/src/views/OpenImage.vue
```diff
@@ -1,13 +1,129 @@
...
+var jquery = require('jquery')
```
louisgv: I think usage of jquery should be avoided. According to the code bellow, it seems like it was used to implement a promise. We should be able to use the native browser promise API instead of importing entire JQuery

interface/frontend/src/views/OpenImage.vue
```diff
@@ -1,13 +1,129 @@
...
+        this.loadImage(this.resolveDICOM)
+      }
+    }
+  },
+  mounted: function () {
+    this.fetchData(this.dicomUrl)
+  },
+  methods: {
+    fetchData (id) {
+      this.$http.get('/api/images/metadata?dicom_location=/images/' + id.slice(this.csName.length + 3))
```
louisgv: To align with the other component, it's best to use the axios API instead.

interface/frontend/src/views/OpenImage.vue
```diff
@@ -1,13 +1,129 @@
 <template>
-  <image-series></image-series>
+  <div class="container">
+    <div class="col-xs-6">
+      <div style="width:512px;height:512px;position:relative;display:inline-block;color:white;"
```
louisgv: Style should be moved to the style tag

jjjmm:
Hey @vessemer. Could you next time indicate that you are working on the issue. You and me were making the same thing on the same time

interface/frontend/src/components/open-image/OpenDICOM.vue
```diff
@@ -0,0 +1,132 @@
+<template>
+  <div class="container">
+    <div class="col-xs-6">
+      <div class="DICOM-container">
+        <div class="DICOM" ref="DICOM"></div>
+      </div>
+      <p>{{ dicom.imageId}}</p>>
```
lamby: spacing here...

interface/frontend/src/components/open-image/OpenDICOM.vue
```diff
@@ -0,0 +1,132 @@
...
+          columns: 512,
+          height: 512,
+          width: 512,
+          color: false,
+          base64data: '',
+          columnPixelSpacing: 1,
+          rowPixelSpacing: 1,
+          sizeInBytes: 512 * 512 * 2
+        },
+        dicomUrl: 'LIDC://LIDC-IDRI-0001/1.3.6.1.4.1.14519.5.2.1.6279.6001.298806137288633453246975630178/1.3.6.1.4.1.14519.5.2.1.6279.6001.179049373636438705059720603192/-95.000000.dcm',
```
lamby: Ooh, magic string!

interface/frontend/src/components/open-image/OpenDICOM.vue
```diff
@@ -0,0 +1,132 @@
...
+          this.loadImage(this.resolveDICOM)
+        }
+      }
+    },
+    mounted: function () {
+      this.fetchData(this.dicomUrl)
+    },
+    methods: {
+      fetchData (id) {
+        this.$axios.get('/api/images/metadata?dicom_location=/images/' + id.slice(this.csName.length + 3))
```
lamby: Can we not let Django generate the URIs instead?

interface/frontend/src/components/open-image/OpenDICOM.vue
vessemer: it was just for testing :)

interface/frontend/src/components/open-image/OpenDICOM.vue
vessemer: Not a really necessary line :)

interface/frontend/src/views/OpenImage.vue
vessemer: Implemented [here](https://github.com/concept-to-clinic/concept-to-clinic/pull/188/commits/fa9eee3ff7707202cbdafd60d263e1661cd1b1cf#diff-70d179422138a3042c2135a282065626R59)

interface/frontend/src/components/open-image/OpenDICOM.vue
vessemer: Sorry, not sure that I understand it correctly: did you mean that this URI prefix should be computed by the Django and then provided to vue?

vessemer:
Hey, @antonow, thanks, it looks much better now
I spent last sleepless nights working on this PR, the main difficulties were related with the [`cornerstoneTool`](https://github.com/chafey/cornerstoneTools), besides for it has highly unstable behaviour (coverage status is 4% only) it also crashes with each update of the dependent [`cornerstone-core`](https://github.com/chafey/cornerstone). The latest versions sustainable for our setup are: 
```javascript
"cornerstone-core": " 0.13.0",
"cornerstone-tools": "0.9.1"
``` 
Relating events implementation: It seems that every guide about `$on` and `$emit` on the internet is wrong (at least for versions described above) so that the decision was made to use the bus to handle events. Here is the current view of DICOM slice selection:
peek 2017-10-27 13-24

interface/frontend/src/views/OpenImage.vue
vessemer: Unfortunately, it fails if only I try one of following:
```javascript
import { cornerstone }  from 'cornerstone-core'
------------------------------------------------------------
import cornerstone  from 'cornerstone-core'
Object.definePrototype(Vue.prototype, '$cornerstone', { value: cornerstone })
------------------------------------------------------------
import cornerstone  from 'cornerstone-core'
Vue.prototype.$cornerstone = cornerstone 
------------------------------------------------------------
import cornerstone from 'cornerstone'

export default {
  install: function(Vue) {
    Object.defineProperty(Vue.prototype, '$cornerstone', { value: cornerstone })
  }
}
```

interface/frontend/src/views/OpenImage.vue
vessemer: done, was used as a first through, thank you for the review!

interface/frontend/src/components/open-image/OpenDICOM.vue
```diff
@@ -0,0 +1,135 @@
...
+        for (var i = 0, strLen = str.length; i < strLen; i += 2) {
+          var lower = str.charCodeAt(i)
+          var upper = str.charCodeAt(i + 1)
+          bufView[index] = lower + (upper << 8)
+          index++
+        }
+        return bufView
+      },
+      getPixelData () {
+        var pixelDataAsString = window.atob(this.dicom.base64data)
```
louisgv: Suggest avoid var in these instances as well. Recommendation is to try to use `const` as much as possible so that we can adhere to immutable functional programming. :smiling_imp: 

Tho, direct alternative for `var` which is scope-safe would be to use `let` :grin:

vessemer: Oh, yeah, thank you for pointing it out :)

interface/frontend/src/components/open-image/OpenDICOM.vue
```diff
@@ -0,0 +1,133 @@
+<template>
+  <div class="DICOM-container">
+    <div class="DICOM" ref="DICOM"></div>
+  </div>
+</template>
+
+<script>
+  import { EventBus } from '../../main.js'
+  const cornerstone = require('cornerstone-core')
+  const Q = require('q')
```
louisgv: `q` is not very well maintained these day and it is also very old. I would suggest either looking forward with the Promise API implemented within the browser itself: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

Or use BlueBird for the perf boost as well as the issue discussed in this article: https://alexn.org/blog/2017/10/11/javascript-promise-leaks-memory.html

Blue bird: http://bluebirdjs.com/docs/getting-started.html

interface/frontend/src/components/open-image/OpenDICOM.vue
```diff
@@ -0,0 +1,133 @@
...
+
+  export default {
+    name: 'open-dicom',
+    data () {
+      return {
+        dicom: {
```
louisgv: I just went through cornerstone's example and have a better understanding of the library. Since we are using vuejs, and since we want this component to be re-usable, it is crucial that we might want to construct this entire object upon rendering the image instead of having a mocked version of it here with some binding to other places to get data.

My suggested implementation looks something like this:

+ This vue comp will accept parameters from parent component, which is the path to the image
+ With that, we move `dicom` to computed and simply do all the fetch and calculation there. With this we might not need to `watch` at all
+ On the same line of thought, we can simply construct the dicom object everytime the path changed, and then apply it using cornerstone.

The above should reduce the intertwine between static data and computed data that we are having here.

louisgv: Refer to this for computed variable in vue: https://vuejs.org/v2/guide/computed.html

lamby:
@louisgv Thanks for the review!

interface/frontend/src/components/open-image/OpenDICOM.vue
vessemer: I've designed data flow through `watch` in order to ensure that data has been uploaded by asynchronous `axios.get`, now for the same goal, I've rewritten it in a promise-based style (with a chain of promises). It also turned out that each click a new canvas instance was created, this unexpected behaviour is fixed now.

interface/frontend/src/components/open-image/OpenDICOM.vue
vessemer: Good point, done by [this](https://github.com/concept-to-clinic/concept-to-clinic/pull/188/commits/f8347df03addd788ca16224f3db246d20c7ad93c).

interface/frontend/src/components/open-image/OpenDICOM.vue
```diff
@@ -0,0 +1,122 @@
...
+            windowCenter: 110,
+            windowWidth: 100,
+            render: cornerstone.renderGrayscaleImage,
+            getPixelData: this.getPixelData,
+            color: false
+          }
+        })
+      },
+      display () {
+        return this.dicom.then((dicom) => {
```
louisgv: For fancy async/await (it's shipped on latest chromium), we can also do something like this:

```
async display () {
   const element = this.$refs.DICOM
   this.initCS(element)
          
   cornerstone.registerImageLoader(this.view.type, this.dicom)
         
    const image = await cornerstone.loadImage(dicom.imageId)

    cornerstone.displayImage(element, image) 
}
```

Probably need to experiment with it a lil bit more. But overall, LGTM!

interface/frontend/src/components/open-image/OpenDICOM.vue
```diff
@@ -0,0 +1,122 @@
...
+      }
+    },
+    data () {
+      return {
+        base64data: null
+      }
+    },
+    computed: {
+      info () {
+        return this.$axios.get(this.view.prefixUrl + this.view.path)
```
louisgv: For reference, here's async/await syntax for this:
```
async info() {
     const response = await this.$axios.get(this.view.prefixUrl + this.view.path); 
     return response.data;
}
```

interface/frontend/src/components/open-image/OpenDICOM.vue
vessemer: Thanks, I've read about it, and implement by [this](https://github.com/concept-to-clinic/concept-to-clinic/pull/188/commits/64484a2ed46aedb1d3fdc5ff58c0cdd344289c6a).

interface/frontend/src/components/open-image/OpenDICOM.vue
vessemer: Thanks, done [here](https://github.com/concept-to-clinic/concept-to-clinic/pull/188/commits/64484a2ed46aedb1d3fdc5ff58c0cdd344289c6a#diff-70d179422138a3042c2135a282065626R30)

interface/frontend/src/components/open-image/OpenDICOM.vue
louisgv: Sweet :ok_hand:

lamby:
Awesome. Thanks guys!

PR #190, opened 2017-10-26
Organize components by view to which they belong

antonow:
## Description
I moved each component except for the nav component, which is used in each view, into a folder named after the view in which the component is used.

There is no official issue related to this.

## Motivation and Context
The components folder already has a lot of files in it. It would be good to establish some patterns for organizing frontend files early on.

## How Has This Been Tested?
I ran `docker-compose -f local.yml up` and saw that the frontend interface still worked the way it was working before.

## CLA


interface/frontend/build/webpack.base.conf.js
```diff
@@ -22,7 +22,7 @@ module.exports = {
     extensions: ['.js', '.vue', '.json'],
     alias: {
       'vue$': 'vue/dist/vue.esm.js',
-      '@': resolve('src'),
+      '@': resolve('src')
```
lamby: Please don't make changes not mentioned in the commit message :)

lamby:
Great stuff! Thanks and merged!

interface/frontend/build/webpack.base.conf.js
antonow: Ok, I will keep that in mind for next time.

Issue #159, opened 2017-10-15
Make the docker-compose documentation service reuse the caches from prediction and interface


reubano:
## Overview
Currently, when running `docker-compose -f local.yml build`, the `documentation` service ignores the caches created by the `prediction` and `interface` services. This results in re-downloading all the python requirements.

## Expected Behavior
The `documentation` service should use the available caches and not re-download any python requirements.

## Technical details

- [Relevant `docker-compose` issue](https://github.com/docker/compose/issues/963) (also includes a few suggestions)

- [Another issue](https://github.com/docker/compose/issues/2828) which displays the desired behavior

- This feature should be implemented in the [`compose/documentation/Dockerfile-dev`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/compose/documentation/Dockerfile-dev), [`compose/prediction/Dockerfile-dev`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/compose/prediction/Dockerfile-dev), [`compose/interface/Dockerfile-dev`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/compose/interface/Dockerfile-dev), and [`local.yml`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/local.yml) files.

## Acceptance criteria

 - [ ] when running `docker-compose -f local.yml build` on a clean system, the python requirements files should not be downloaded by `documentation` service. 
 - [ ] when running `docker-compose -f local.yml build` on a clean system, the `documentation` service should use the caches provided by the `prediction` and `interface` services.

*NOTE: All PRs must follow the standard PR checklist.*

WGierke:
A friend of mine proposes that we should build our own base image which is based on the union of the pip requirements. Those, the base image is only supposed to be compiled once. If services use different versions of a library, they have to install them themselves after the base image was created. Just throwing in my two cents :)

PR #199, opened 2017-11-02
#159 Docker: Reuse Caches

WGierke:
First, a base image is built that installs all needed pip requirements. The docker containers, that need parts of the requirements (prediction, interface and documentation services), then are built on top of that base image. Thus, building the documentation service doesn't require to install all pip requirements again resulting in a great speed-up of the container building.

This addresses #159 

## How Has This Been Tested?
`docker-compose -f local.yml build && docker-compose -f local.yml up` builds all containers properly and fires them up.

## CLA

PS: Accelerated the build time on Travis from μ=23.5min ± 2.2 min (measured on the last 25 builds) to 18 min (first successful build of this branch) :tada: The time it takes to build the services locally should thus also greatly be decreased.

### DIFF
--- a/compose/base/Dockerfile-dev
+++ b/compose/base/Dockerfile-dev
@@ -1,0 +1,26 @@
+FROM ubuntu:rolling
+ENV PYTHONUNBUFFERED 1
+RUN apt-get update && apt-get install -y tcl tk python3.6 python3.6-tk wget python-opencv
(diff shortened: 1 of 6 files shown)


lamby:
> building the documentation service doesn't require to install all pip requirements again 

As someone who accidentally needed to remake these images yesterday on a terrible internet connection, oh my gosh thank you ♥

Issue #203, opened 2017-11-02
ModuleNotFoundError: No module named 'src'


lamby:
## Current Behavior

```
Traceback (most recent call last):
  File "/app/src/factory.py", line 53, in <module>
    app = create_app()
  File "/app/src/factory.py", line 36, in create_app
    from .views import blueprint
  File "/app/src/views.py", line 9, in <module>
    from .algorithms.classify import trained_model as classify_trained_model
  File "/app/src/algorithms/classify/trained_model.py", line 10, in <module>
    from src.algorithms.classify.src import gtr123_model
ModuleNotFoundError: No module named 'src'
```

## Steps to Reproduce
1. Visit http://localhost:8001/
2. See traceback

## Checklist before submitting
- [ ] I have confirmed this using the officially supported Docker Compose setup using the `local.yml` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [ ] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [ ] I provided a **minimal code snippet** or list of steps that reproduces the bug.
- [ ] I provided **screenshots** where appropriate
- [ ] I filled out all the relevant sections of this template

reubano:
FYI, here's how it was [previously fixed](https://github.com/concept-to-clinic/concept-to-clinic/commit/a482fd890af5ef81c088464f619c8a427e2199ea)

Issue #160, opened 2017-10-15
Update travis tests to catch route errors


reubano:
## Overview
Currently, the travis test passes even if some routes are non-functional.

E.g.

- `git checkout 317b8c18`
- `docker-compose -f local.yml build`
- `sh tests/test_docker.sh`  # tests should pass
- `docker-compose -f local.yml up`
- `open http://0.0.0.0:8001/`  # page should error

## Expected Behavior
Travis should test all navigable routes and ensure that they properly function.

## Technical details

At a minimum, the following home pages should be tested:

- http://0.0.0.0:8000/
- http://0.0.0.0:8001/
- http://0.0.0.0:8080/

The appropriate routes for each home page should be tested as well.

e.g., `http://0.0.0.0:8000/api` and `http://0.0.0.0:8001/classify/predict`

- This feature should be implemented in the [`tests/test_docker.sh`](https://github.com/concept-to-clinic/concept-to-clinic/tree/master/tests/test_docker.sh) file.

## Acceptance criteria

 - [ ] The travis test should fail if any of the service routes produces an error. 

*NOTE: All PRs must follow the standard PR checklist.*

adrianjoseph:
Hi Reubano, 
What are you looking for when you say that the pages should be "tested"? In other words, what qualifies as "tested"? Also, what is a "travis test"

From what I can see:
http://0.0.0.0:8000/ - Has a 404 page not found error
http://0.0.0.0:8001/ - Gets a "ERR_CONNECTION_REFUSED" error
http://0.0.0.0:8080/ - This page loads and it seems you can see a drill down of a set of related fil

As an fyi: I am trying to get my feet wet with this process, so some of the asks may seem basic, but I am just trying to understand how to view and make changes where applicable.

reubano:
Hi @adrianjoseph [these are the travis tests](https://travis-ci.org/concept-to-clinic/concept-to-clinic). It runs the tests [in this file](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/tests/test_docker.sh). So by "the pages should be tested", I mean that if you checkout commit `317b8c18`, the travis tests should fail since not all the routes work. Does that make sense?

isms:
I'm still not clear on why we would want a gigantic integration test instead of testing the services individually?

reubano:
[...]

PR #217, opened 2017-11-08
#203 Fix error showing no module named 'src' found.

musale:
Closes the issue #203 that threw an error whe you visit the http://localhost:8001/ endpoint in the prediction application.

## Description
Using `python -m flask run ...` to start server. More info in docs [here](http://flask.pocoo.org/docs/0.11/cli/#basic-usage)

This closes issue #203 
### Steps to reproduce
- `docker-compose -f local.yml up`
- visit [http://localhost:8001/](http://localhost:8001)
- Should load the home page

## Motivation and Context
This removes the traceback error when you visit [http://localhost:8001/](http://localhost:8001)

## How Has This Been Tested?
When you visit [http://localhost:8001/](http://localhost:8001) you should not get an error

## CLA

### DIFF
--- a/local.yml
+++ b/local.yml
@@ -64,7 +64,7 @@ services:
     build:
       context: .
       dockerfile: ./compose/prediction/Dockerfile-dev
-    command: flask run --host=0.0.0.0 --port=8001
+    command: python -m flask run --host=0.0.0.0 --port=8001
     depends_on:
       - base
     environment:


WGierke:
Would it make sense to write a test that checks whether the traceback still exists when visiting the page? This would enable us to notice when the page breaks again.

reubano:
@WGierke see #160

WGierke:
Nice. Thanks and sorry :)

lamby:
Thanks!

PR #218, opened 2017-11-08
#160 Add Route Testing

WGierke:
I added a simple Python script that can test whether routes error. It is highly customizable (e.g. one could add to check for the occurrence of the word "error" or "exception" etc.).

This addresses #160 

## Motivation and Context
Currently, the Travis test passes even if some routes are non-functional (like http://0.0.0.0:8001/).

## How Has This Been Tested?
It's currently not passing due to the unfixed issue #203 .
```bash
➜  concept-to-clinic git:(160_route_errors) ✗ bash tests/test_docker.sh
...
+ python tests/test_routes.py
http://localhost:8001/ returns 500
+ echo 'ERROR: test_routes did not pass. Check above for details.'
ERROR: test_routes did not pass. Check above for details.
+ exit 1
```

## CLA

### DIFF
--- a/tests/test_routes.py
+++ b/tests/test_routes.py
@@ -1,12 +1,15 @@
 import sys
+import time
+
 import requests
 
+TIMEOUT = 180
 LOCALHOST = "http://localhost:{}/"
-SERVICE_PORTS = {"INTERFACE":     {"PORT": 8000, "SITES": ["api"]},
-                 "VUE":           {"PORT": 8080, "SITES": []},
-                 "PREDICTION":    {"PORT": 8001, "SITES": ["classify/predict/",
-                                                           "identify/predict/",
-                                                           "segment/predict/"]},
+SERVICE_PORTS = {"INTERFACE": {"PORT": 8000, "SITES": ["api"]},
+                 "VUE": {"PORT": 8080, "SITES": []},
+                 "PREDICTION": {"PORT": 8001, "SITES": ["classify/predict/",
+                                                        "identify/predict/",
+                                                        "segment/predict/"]},
                  "DOCUMENTATION": {"PORT": 8002, "SITES": []}}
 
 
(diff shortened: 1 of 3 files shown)


tests/test_routes.py
```diff
@@ -0,0 +1,22 @@
...
+
+LOCALHOST = "http://localhost:{}/"
+SERVICE_PORTS = {"INTERFACE":    {"PORT": 8000, "SITES": ["api"]},
+                 "VUE":          {"PORT": 8080, "SITES": []},
+                 "PREDICTION":   {"PORT": 8001, "SITES": ["classify/segment"]},
+                 "DOCUMENTATION":{"PORT": 8002, "SITES": []}}
+
+def check_page(url):
+    response = requests.get(url)
+    if response.status_code != 200:
```
lamby: The canonical way of doing this with ``requests`` is using/catching ``raise_for_status`` - I think it gives you a nicer error message, but I'm not sure. Can you have a look?

WGierke:
@lamby Thanks for the hint, that indeed is a way better solution to test for unsuccessful responses. However, apparently one has to do some more bash magic than expected. The docker containers must be started so the ports are served at localhost. However, the `compile_docs` service is producing such a big output that it hits the Travis log limit of 4MB and the test fails (apart from that, fixing `compile_docs` might be a new separate issue?). Thus, I'm piping the output to /dev/null. While the tests now work on my machine, they're still failing on Travis:
```bash
➜  concept-to-clinic git:(160_route_errors) ✗ git log | head -n 1
commit 0e8525940f3fda572745e400f8cb1fe01b35c329
➜  concept-to-clinic git:(160_route_errors) ✗ bash tests/test_docker.sh
...
+ docker-compose -f local.yml run interface python manage.py makemigrations --dry-run --check
Starting concepttoclinic_postgres_1 ... 
Starting base ... 
Starting base ... done
Starting concepttoclinic_prediction_1 ... done
Postgres is up - continuing...
/usr/local/lib/python3.6/dist-packages/environ/environ.py:618: UserWarning: Error reading /app/.env - if you're not configuring your environment separately, check this.
  "environment separately, check this." % env_file)
No changes detected
+ sleep 60
+ docker-compose -f local.yml up
+ python tests/test_routes.py
Fetching http://localhost:8000/ ...
Fetching http://localhost:8000/api ...
Fetching http://localhost:8080/ ...
Fetching http://localhost:8001/ ...
Fetching http://localhost:8001/classify/predict/ ...
Fetching http://localhost:8001/identify/predict/ ...
Fetching http://localhost:8001/segment/predict/ ...
Fetching http://localhost:8002/ ...
➜  concept-to-clinic git:(160_route_errors) ✗ echo "$?"        
0
```
I'd be happy over any solution suggestions :)

lamby:
I really don't like the sleep - can we not test instead? :p

WGierke:
:D Okay so how about, instead of `sleep 60`, we try to connect to the 4 ports and after 60 seconds we time out and exit with 1, otherwise we perform the tests?

lamby:
> are ready to be tested

Right, can't we just wait for that to happen by testing them? Not sure why we need a timeout for this.

WGierke:
Okay but what if a port won't be opened because the service died for some reason? In that case we test endlessly which might be no problem on Travis since there is a timeout but not on local computers.

lamby:
@WGierke Sorry for the delay in getting back to you. Perhaps I'm misunderstanding something - if the port is open and subsequently closes, we are safe to assumine something is FUBAR and fail the test. However, AIUI this is about waiting for things to spin-up, no? Sorry if I'm misunderstanding something here. :)

WGierke:
` this is about waiting for things to spin-up, no`
Exactly, that's what I meant. Sorry for the confusion :)

lamby:
Getcha. So to summarise, I understand the need to timeout in the case that the port never materialises, but it would be still be great to *start* the tests before this 60 seconds if the ports are open. ie. would you be okay pushing ahead with https://github.com/concept-to-clinic/concept-to-clinic/pull/218#issuecomment-343729340 ? :)

WGierke:
@lamby Yes ;) It should be done now.

lamby:
Awesome, thanks :)

Issue #198, opened 2017-11-01
Correct case of links in documentation sidebar


lamby:
On: https://concept-to-clinic.readthedocs.io/

## Current Behavior
[](https://i.imgur.com/djeE94L.jpg) (https://i.imgur.com/djeE94L.jpg)

## Expected Behavior
"Interface" and "Prediction" (ie. title case, not all lower case)

## Steps to Reproduce

1. https://concept-to-clinic.readthedocs.io
2. Scroll to bottom of navigation

## Checklist before submitting
- [ ] I have confirmed this using the officially supported Docker Compose setup using the `local.yml` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [ ] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [ ] I provided a **minimal code snippet** or list of steps that reproduces the bug.
- [ ] I provided **screenshots** where appropriate
- [ ] I filled out all the relevant sections of this template

WGierke:
( :+1: for the first-pr-only label)

PR #220, opened 2017-11-10
#198: Update case of links in documentation sidebar

musale:
## Description
Create new files to handle the index pages of the `prediction` and `interface` modules inside `docs`.
Generally, `sphinx-api` on these two modules auto generates the documentation. Ignore the previously generated `modules.rst` files because they're automatically generated and changes made to them are overridden.

Fixes issue #198.

## Motivation and Context
Makes the cases of the `Code Documentation` titles begin with upper case.

## How Has This Been Tested?
- `docker-compose -f local.yml build && docker-compose -f local.yml up`
- Open [http://localhost:8002/](http://localhost:8002/)
- Titles in the section *Code Documention* should start with upper case

## CLA

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -86,7 +86,9 @@ language = None
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This patterns also effect to html_static_path and html_extra_path
-exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+exclude_patterns = [
+    '_build', 'Thumbs.db', '.DS_Store', 'apidoc_interface/modules.rst',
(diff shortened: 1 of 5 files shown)


docs/conf.py
```diff
@@ -86,7 +86,9 @@
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This patterns also effect to html_static_path and html_extra_path
-exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+exclude_patterns = [
```
lamby: I wonder how sane it would be to parse `.gitignore` :)

lamby:
LGTM :)

PR #221, opened 2017-11-10
Update PR TEMPLATE to include prediction algorithm metrics

reubano:
## Description

This PR updates `PULL_REQUEST_TEMPLATE.md` to include various prediction algorithm metrics such as logloss and prediction time. View [formatted file](https://github.com/concept-to-clinic/concept-to-clinic/blob/630e1aeae7673b2bb0fc466c99f7612047438cf7/.github/PULL_REQUEST_TEMPLATE.md#metrics-if-appropriate).

## Motivation and Context

This PR will allow us to evaluate current and future machine learning algorithms in a standardized and objective fashion.

## CLA
- [ x ] I have signed the CLA; if other committers are in the commit history, they have signed the CLA as well

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -21,5 +21,52 @@
 ## Screenshots (if appropriate):
 
 
+## Metrics (if appropriate):
+
+If you submitting a PR for a prediction algorithm (segmentation, identification,
+or classification) please fill in values for as many as the below statistics as
+are relevant.
+
+*algorithms by metric*
+
+metric | relevant algorithms
+-------|--------------------
+[accuracy <sup>1</sup> <sup>2</sup>](https://stats.stackexchange.com/a/231237/143678) | classification, identification
+[data IO](https://unix.stackexchange.com/questions/55212) | classification, identification, segmentation
+[Dice coefficient <sup>3</sup>](https://en.wikipedia.org/wiki/S%C3%B8rensen%E2%80%93Dice_coefficient) | segmentation
+[disk space usage](https://www.cyberciti.biz/faq/linux-check-disk-space-command) | classification, identification, segmentation
+[Hausdorff distance <sup>3</sup>](https://en.wikipedia.org/wiki/Hausdorff_distance) | segmentation
+[Jaccard index](https://en.wikipedia.org/wiki/Jaccard_index) | segmentation
+[Log Loss](http://wiki.fast.ai/index.php/Log_Loss) | classification, identification <sup>4</sup>
+[memory usage](https://stackoverflow.com/questions/110259) | classification, identification, segmentation
+[prediction time <sup>2</sup>](https://stackoverflow.com/questions/385408) | classification, identification, segmentation
+[sensitivity <sup>3</sup>](http://wiki.fast.ai/index.php/Deep_Learning_Glossary#Recall) | segmentation
+[specificity <sup>3</sup>](http://wiki.fast.ai/index.php/Deep_Learning_Glossary#Specificity) | segmentation
+[training time <sup>2</sup>](https://stackoverflow.com/questions/385408) | classification, identification, segmentation
+
+*notes*
+
+1. Use 5-fold cross validation if there is enough time and computational power available, otherwise use a holdout set
+1. This metric may be automatically calculated by the machine learning architecture, e.g., Keras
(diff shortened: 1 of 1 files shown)


lamby:
LGTM

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -21,5 +21,37 @@
 ## Screenshots (if appropriate):
 
 
+## Metrics (if appropriate):
+
+<!--- 
+If you submitting a PR for a prediction algorithm (segmentation, identification, or classification
```
WGierke: Would you mind inserting a closing parenthesis at the end? :)

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -21,5 +21,37 @@
...
+logloss          | https://www.kaggle.com/c/data-science-bowl-2017#evaluation    | identification, classification
+training time    | https://stackoverflow.com/questions/385408                    | segmentation, identification, classification
+prediction time  | https://stackoverflow.com/questions/385408                    | segmentation, identification, classification
+data IO          | https://unix.stackexchange.com/questions/55212                | segmentation, identification, classification
+disk space usage | https://www.cyberciti.biz/faq/linux-check-disk-space-command/ | segmentation, identification, classification
+memory usage     | https://stackoverflow.com/questions/110259                    | segmentation, identification, classification
+
+algorithm      | relevant metrics
+---------------|------------------
+segmentation   | jaccard loss, training time, prediction time, data IO, disk space usage, memory usage
```
WGierke: How about adding some metrics for a segmentation algorithm that have already been implemented to provide more evaluation information such as [Hausdorff distance, sensitivity, specificity or dice coefficient](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/prediction/src/algorithms/segment/src/evaluate.py)?

reubano:
One metric I left out is an estimation of [model over/under fitting](http://wiki.fast.ai/index.php/Lesson_3_Notes#On_Deeper_Finetuning) ([video](https://www.youtube.com/watch?v=6kwQEBMandw)). This is something we should probably address at some point.

lamby:
@reubano Did you address all of @WGierke's comments? :)

reubano:
@WGierke yep

lamby:
LGTM.. :)

WGierke:
@reubano one thing that just came to my mind: isn't identification about predicting a (numerical) location of a nodule? How should we then calculate the accuracy based on that? From my understanding the accuracy can only be used for classification problems, since here we have a regression problem something like the mean squared error would be more appropriate, wouldn't it? Or do you want to put every x, y and z value in their own classes? This way the accuracy wouldn't differ for a prediction in which e.g. x_hat is x_true+1 and a prediction where x_hat is x_true+100.

WGierke:
I consulted my pillow and it said that one could handle an identification as correct if the predicted nodule location is within the boundaries of the nodule, but not necessary the centroid (for LIDC images we can get the boundaries easily using [pylidc](https://github.com/pylidc/pylidc/blob/ea04c829a50e3598934e3c2f603e20969fe8381f/pylidc/Annotation.py#L227). Another way to achieve this would be to handle an identification as correct if the predicted locations are within a Ɛ-pipe of the true location (x_true - Ɛ <=  x_hat <= x_true + Ɛ, y_true - Ɛ <= y_hat ...). Ɛ could vary for z since the size of this dimension differs from x and y.
This brings me to my third point: the accuracy is the ratio of correct classifications from all the classifications the algorithm performed. If the algorithm only predicts one nodule out of five and the prediction is correct, its accuracy would be 100% but it still wasn't very helpful since it missed four nodules. A metric to cover this case would be the [recall](https://en.wikipedia.org/wiki/Precision_and_recall) which can also easily be computed. Any thoughts on that? :)

reubano:
@WGierke good points! I think one way to address the identification's accuracy calculation is to do something similar as mentioned in the log loss note:

> In order to calculate Log Loss for identification, the data needs to be arranged in a way that shows for each pixel, whether or not it is a nodule centriod. Restated, the pixel level labels of 1/0 would correspond to centriod/not-centriod.

This would convert identification into something more classification like. As to Ɛ error value you suggested, I think that's a great idea. We could implement that by counting a label as being matched if it is `x` number of pixels away from the true centroid.

If I understand your 3rd point correctly, you are saying accuracy only accounts for `precision`. The recall metric you referenced is also known as `sensitivity`. Since that metric is already included in the list, I agree we should just add identification as a relevant algorithm.

Issue #206, opened 2017-11-02
IsADirectoryError: [Errno 21] Is a directory: '/'


lamby:
```
interface_1      |   File "/usr/local/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
interface_1      |     return view_func(*args, **kwargs)
interface_1      |   File "/usr/local/lib/python3.6/site-packages/django/views/generic/base.py", line 68, in view
interface_1      |     return self.dispatch(request, *args, **kwargs)
interface_1      |   File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 489, in dispatch
interface_1      |     response = self.handle_exception(exc)
interface_1      |   File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 449, in handle_exception
interface_1      |     self.raise_uncaught_exception(exc)
interface_1      |   File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 486, in dispatch
interface_1      |     response = handler(request, *args, **kwargs)
interface_1      |   File "/app/backend/api/views.py", line 58, in get
interface_1      |     ds = dicom.read_file(path, force=True)
interface_1      |   File "/usr/local/lib/python3.6/site-packages/dicom/filereader.py", line 589, in read_file
interface_1      |     fp = open(fp, 'rb')
interface_1      | IsADirectoryError: [Errno 21] Is a directory: '/'
```

## Steps to Reproduce
1. http://localhost:8080/#/
2. See traceback in console

## Checklist before submitting
- [ ] I have confirmed this using the officially supported Docker Compose setup using the `local.yml` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [ ] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [ ] I provided a **minimal code snippet** or list of steps that reproduces the bug.
- [ ] I provided **screenshots** where appropriate
- [ ] I filled out all the relevant sections of this template

PR #224, opened 2017-11-14
#206 IsADirectoryError fix

musale:
Loading [http://localhost:8080/](http://localhost:8080) throws an `IsADirectoryError: [Errno 21] Is a directory: '/'`

## Description
The endpoint is called every time the front end loads. Since the `dicom.readfile()` method is an IO op, I do a `try/except` to catch the `IOError` 21. As it is, this error is only anticipated on page load because there's no default path to a dicom file. I return an 'empty' response back so that the error is handled on the server-side and presents the client side an opportunity to handle it too.

### To handle the empty response from the server.

`Cornerstone` throws errors if handed empty fields on the front end. Since the functions trigger calls to the server endpoint on load (I am assuming this is because of the `async/await` computed functions -- please correct my reasoning here if I'm wrong) and has nothing to display, catering for instances of empty values mitigates these unnecessary errors. Check the changes in [OpenDICOM.vue file](https://github.com/musale/concept-to-clinic/blob/fb39d63d387f2706c6e69792c3739bf61f76506f/interface/frontend/src/components/open-image/OpenDICOM.vue).

Fixes issue #206 

## Motivation and Context
This change removes the error logs for `IsADirectoryError` and also handles instances where a dicom file is not found.

## How Has This Been Tested?
- Load [http://localhost:8080/](http://localhost:8080)
- There should be no error traceback

## CLA

### DIFF
--- a/interface/backend/api/views.py
+++ b/interface/backend/api/views.py
@@ -55,7 +55,11 @@ class ImageMetadataApiView(APIView):
             type: string
         '''
         path = request.GET['dicom_location']
-        ds = dicom.read_file(path, force=True)
+        try:
+            ds = dicom.read_file(path, force=True)
+        except IOError as err:
+            print(err)
+            return Response(serializers.DicomMetadataSerializer().data)
         return Response(serializers.DicomMetadataSerializer(ds).data)
 
 
--- a/interface/frontend/build/dev-server.js
+++ b/interface/frontend/build/dev-server.js
@@ -25,7 +25,11 @@ var compiler = webpack(webpackConfig)
 
 var devMiddleware = require('webpack-dev-middleware')(compiler, {
   publicPath: webpackConfig.output.publicPath,
-  quiet: true
+  quiet: true,
+  watchOptions: {
+    aggregateTimeout: 300,
+    poll: 200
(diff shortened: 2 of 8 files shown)


lamby:
Hi @musale! Did you mean to squash these into one commit? If you still have the separate changes on your local branch it would be great if you could push them instead; I'm seeing a bunch of whitespace changes that is making the rest hard to review in terms of what it fixes! No worries if not, but I thought I would just ping first instead of picking it apart manually :)

musale:
hi @lamby yes. I meant to squash it into one commit.

musale:
@lamby I think the update should be clearer on the changed files now 😄

interface/backend/api/views.py
```diff
@@ -55,7 +55,11 @@ def get(self, request):
             type: string
         '''
         path = request.GET['dicom_location']
-        ds = dicom.read_file(path, force=True)
+        try:
+            ds = dicom.read_file(path, force=True)
+        except IOError as err:
+            print(err)
```
lamby: Hm, we shouldn't be `print`ing during a view method, no? :)

Issue #205, opened 2017-11-02
Export PDF functionality


lamby:
![](https://i.imgur.com/qzuzIKD.jpg)

## Expected Behavior
Pressing *Export* exports the page as a PDF

## Current Behavior
(Not implemented)

## Checklist before submitting
- [ ] I have confirmed this using the officially supported Docker Compose setup using the `local.yml` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [ ] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [ ] I provided a **minimal code snippet** or list of steps that reproduces the bug.
- [ ] I provided **screenshots** where appropriate
- [ ] I filled out all the relevant sections of this template

Serhiy-Shekhovtsov:
@lamby,  I can see the **interface-backend** tag, is there a particular reason why this has to be done on the server side? I am asking this, cause implementing it on the client side should be an easy task.

lamby:
@Serhiy-Shekhovtsov Implementing client-side is fine :)

Issue #215, opened 2017-11-08
Eliminate/Restrict Jquery use for cornerstone-tools


louisgv:
## Expected Behavior

Related to #197 per @vessemer request.

The front-end should not be depending on JQuery as a dependence, unless there is popular demand from contributor. Per #197, it is being imported due to another dependency for advanced medical image processing called `cornerstone-tool`.

It is desirable if we can devise a minimal version of JQuery that has all the required function that CornerStone-tool need, or simply find a way to remove JQuery.

## Current Behavior

The front-end application starting after #197 will be relying on JQuery as a dependency.

## Possible Solution

+ Searching for `external.$` in `cornerstone-tool` [repository](https://github.com/chafey/cornerstoneTools), then log down all method the library is relying on.
+ Extract those method from jQuery and export them into an internal library, with which we expose it to cornerstoneTools
+ Potentially make PR upstream to cornerstone-tool

## Context (Environment)

+ JQuery is a sizable library, however we're already using vue.js and since most of the feature we need from JQuery can easily be implemented with vue, it is necessary to not be depended on jQuery.

## Checklist before submitting
- [ + ] I have confirmed this using the officially supported Docker Compose setup using the `local.yml` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [ + ] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [ + ] I provided a **minimal code snippet** or list of steps that reproduces the bug.
- [ ] I provided **screenshots** where appropriate
- [ + ] I filled out all the relevant sections of this template

truefalse10:
is anyone working on this right now? otherwise i would start

Serhiy-Shekhovtsov:
@truefalse10 btw, looks like they have a branch for this: https://github.com/chafey/cornerstone/tree/zachasme-es6
More details [here](https://github.com/chafey/cornerstone/issues/180) and [here](https://github.com/chafey/cornerstone/issues/92)

dannyrb:
It's worth noting that cornerstone is on track to drop its dependency on jQuery. Work is actively being done. It might be worth waiting for the official change, and then update to latest.

swederik:
We just released Cornerstone Tools 2.0.0 which no longer depends on jQuery. Let us know if you run into any bugs. Thanks!

lamby:
[...]

PR #225, opened 2017-11-14
#205 Export pdf functionality.

musale:
Clicking on the **Report and Export** section, you have an **Export** button that downloads the page as a `.pdf` file.

## Description
Create a pdf file of the section of the site in the **Report and Export** section

Addresses issue #205 

## Motivation and Context
This feature allows downloading of the page information in `.pdf` format.

## How Has This Been Tested?
- Load [http://localhost:8080/](http://localhost:8080/)
- Click on the **Report and Export** menu section
- Click on the **Export** button
- It should download the page information in pdf format

## Output samples
[PDF file with eye candy](https://www.scribd.com/document/364610595/Rsna-Standard-Template-14)
[PDF file without eye candy](https://www.scribd.com/document/364610685/Rsna-Standard-Template-12)
## CLA


interface/frontend/src/components/report-and-export/RSNAStandardTemplate.vue
```diff
@@ -197,6 +197,9 @@ import RSNAStandardTemplateFancy from './RSNAStandardTemplateFancy'
 import OneLineParagraph from './OneLineParagraph'
 
 import Nodule from '../annotate-and-segment/Nodule'
+import html2canvas from 'html2canvas'
+import JSPDF from 'jspdf'
+import $ from 'jquery-slim'
```
vessemer: I guess due to the #215 we have to avoid usage of `jquery`, even slim version.
From the code below, I see that `jquery` was only used to get the desired DOM. 
This can easily be done through `$refs` method from [`Vue API`](https://vuejs.org/v2/api/).

interface/frontend/src/components/report-and-export/RSNAStandardTemplate.vue
```diff
@@ -213,7 +216,27 @@ export default {
   props: [ 'rsna' ],
   methods: {
     exportRSNA () {
-
+      // specified the column with all the 3 components in the view
+      const pdfSection = $('#pdf-section')
+      html2canvas(pdfSection).then(canvas => {
+        var imgData = canvas.toDataURL('image/png')
```
vessemer: Not sure, but think it'll be better to follow the @louisgv suggestion to use `let` and `const` in lue of `var`

interface/frontend/src/components/report-and-export/RSNAStandardTemplate.vue
musale: uhh had missed this one. Let me fix in a few :thumbsup: I am quite new Vue

Serhiy-Shekhovtsov: I am not sure `html2canvas` is a good idea since it will basically make a photo of selected element. So if you make the windows small, the result resolution will be really low. And second thing - it doesn't leave the text as text. Everything inside specified element will be taken as an image.

lamby:
(Updated to match!)

lamby:
I'd love to see an example output? Can you upload something somewhere? :)

musale:
@lamby updated with some sample files

Issue #228, opened 2017-11-16
Any reason for having HyperlinkedModelSerializer instead of ModelSerializer?


Serhiy-Shekhovtsov:
Is there any reason for using `HyperlinkedModelSerializer` for serializing the data before sending it to client side?

### Having this serializer have following drawbacks:

- Returned data doesn't have the **id**. Here is an example of data for `/api/candidates.json` method:
`{"url":"http://interface:8001/api/candidates/1.json","centroid":{ ... },"created":" ... ","probability_concerning":0.3,"case":"http://interface:8001/api/cases/1.json"}`
- Returned URL is not usable. 
  - As you can see, it's absolute and since the port doesn't match the Vue.js server's port - AJAX requests to that URL are blocked by the browser.
  - Also the url contains **.json** extension. For example, services we have for marking/dismissing candidates look like this: `candidates/1/mark`, `candidates/1/dismiss`

### So, what are the benefits?
Maybe I am missing something here, since I am totally new to django, but it looks like replacing `HyperlinkedModelSerializer` with `ModelSerializer` helps to solve the issue of missing **id** and it also doesn't return unusable urls.

## Possible Solution
AFAIK, we can still use `HyperlinkedModelSerializer` but patch it by overriding a method for creating serializer and passing `None` as `request`. Or we can simply use to `ModelSerializer`.

isms:
Hyperlinked fields let the frontend get canonical URLs from the backend rather than doing `domain + id + endpoint` stuff in frontend JS which will immediately break if the endpoints change or if a field other than `id` is used for lookups in the future. Instead the idea is that you can fetch a related resource by fetching `object.url`. You can make it work either way but including links seems to be considered truer to the REST concept.

Two address the two specifics:

1. "Also the url contains .json extension" - is there an example of an issue? Or is this a setting that can simply be changed on the serializer or view? Seems like views are still free to return custom endpoints like `/mark` and `/dismiss` that don't go by this scheme as the only thing HyperlinkedModelSerializer tries to add links for is the specific instance it is serializing.

[...]

PR #231, opened 2017-11-18
Fixed interface API to return relative urls instead of absoulte

Serhiy-Shekhovtsov:
Now each `[Model]ViewSet` will inherit from `ViewSetBase` and that base view class has `get_serializer_context` method overridden to set `'request': None` for the serialization context.

## Description
After this change serialized entities have relative URLs instead of absolute and it makes them usable from the Vue.js application. Problem with absolute URLs - interface API is hosted on different port number then *Vue.js* and requests to that port number being blocked by browser because of CORS policy. Now the URLs are pointing to the same host and will be redirected by *Node.js* server.

Fix suggested by @isms when discussing #228

## CLA

### DIFF
--- a/interface/backend/api/views.py
+++ b/interface/backend/api/views.py
@@ -22,22 +22,32 @@ from rest_framework.views import APIView
 from ..cases import enums
 
 
-class CaseViewSet(viewsets.ModelViewSet):
+class ViewSetBase(viewsets.ModelViewSet):
+    def get_serializer_context(self):
+        context = super(ViewSetBase, self).get_serializer_context()
+
+        # getting rid of absulute URLs
+        context.update({'request': None})
+
+        return context
+
+
+class CaseViewSet(ViewSetBase):
     queryset = Case.objects.all()
     serializer_class = serializers.CaseSerializer
 
 
-class CandidateViewSet(viewsets.ModelViewSet):
+class CandidateViewSet(ViewSetBase):
     queryset = Candidate.objects.all()
     serializer_class = serializers.CandidateSerializer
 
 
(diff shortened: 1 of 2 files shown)


interface/backend/api/views.py
```diff
@@ -22,22 +22,27 @@
 from ..cases import enums
 
 
-class CaseViewSet(viewsets.ModelViewSet):
+class ViewSetBase(viewsets.ModelViewSet):
```
isms: Request this get a more descriptive name like `RequestNoneViewSet`.

interface/backend/api/views.py
```diff
@@ -22,22 +22,27 @@
...
+    def get_serializer_context(self):
+        return {'request': None}
```
isms: Almost certainly we want to get the context first from `super().get_serializer_context()` and then update only the value for the `'request'` key.

isms:
@Serhiy-Shekhovtsov Couple minor changes requested but other than that looks great. Thanks!

interface/backend/api/views.py
Serhiy-Shekhovtsov: Yes, nice catch.

interface/backend/api/views.py
Serhiy-Shekhovtsov: Not sure about this. This class may contain other basic functions in future, so we can call it `ViewSetBase` and put all basic configuration here without fixing all references of it. What do you think?

isms: Fair enough, I think it's fine for now.

isms:
Merged, thanks!

PR #232, opened 2017-11-18
fixed Vue.js startup issue

Serhiy-Shekhovtsov:
Front-end application was not working because of the following error:
`:value conflicts with v-model on the same element because the latter already expands to a value binding internally`

## CLA


lamby:
For someone unfamiliar with Vue.js, can you describe what the issue is here in more detail? :)

Serhiy-Shekhovtsov:
@lamby after checking this again I have found that it's a Vue.js bug and it has been fixed in latest version. Will fix my PR. You can check more details [here](https://github.com/vuejs/vue/issues/7084), but basically it was throwing an error when trying to pass `v-model="..."` and `:value="..."` same time to one element.

Issue #202, opened 2017-11-02
Improve open imagery "Import" button UX


lamby:
## Current Behavior

![](https://i.imgur.com/g7P0jyI.jpg)

Pressing the *Import* button expands the filename browser. This makes sense. However, pressing *Import* again contracts the browser, which does not make much sense as you

## Expected Behavior

Unsure about the ideal behaviour. One possibility at this point is that the copy changes after it has expanded (eg. "hide import"). Alternatively, we could hide the button altogether. Or we explicitly change it to "Show files / hide files". Please experiment!

## Checklist before submitting
- [ ] I have confirmed this using the officially supported Docker Compose setup using the `local.yml` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [ ] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [ ] I provided a **minimal code snippet** or list of steps that reproduces the bug.
- [ ] I provided **screenshots** where appropriate
- [ ] I filled out all the relevant sections of this template

louisgv:
If the allows user to import multiple image, then we should remove the call to contracts the browser from the import button event handler, and instead add it to a drop-down button element similar to what we have with nodule.

Serhiy-Shekhovtsov:
I am not sure if we should allow selecting multiple images. In documentation we can find the following: 
> Allow user to select a given image and start a new “Case.” This image will be used in the identification step next
Also it will complicate some things, including this story: Prevent progress to next stage of analysis until "completed" the current step (#204).
Please, let me know what do you think? (This issue is blocking me with #145)

louisgv:
@Serhiy-Shekhovtsov After reading the requirement again, I think we should separate the two function into 2 UI elements (button), one for importing image and one for expanding the image. So maybe for #145, you can simply change the callback of the import button to do what need to be done, or you can also add a new button to import (to avoid conflict with this issue), and then we can refactor the UI at a later stage?

louisgv:
Hmm, seems like #204 and #202 (this) will be sharing some higher order state. I will do some experiment with this UI element as well as implementing a parent state (to keep track of the step for #204) then :-?

Serhiy-Shekhovtsov:
I am going to do the following: 

- Show if there is any case started. If started - show a message that starting a new case will drop the current one. 
- Show list of images in local system. User can click on image name to view it. (this is already implemented)
[...]

Issue #145, opened 2017-09-30
Import image series


Serhiy-Shekhovtsov:
I can't find how to import images for opening. Is this feature implemented? If it's not - is there a workaround for importing?

## Expected Behavior
User should be able to import images displayed in "Import image series" section.

## Current Behavior
There is tree view for browsing file system, but there is no way to select and import file\folder.
![image](https://user-images.githubusercontent.com/607527/31047402-c4037f9e-a612-11e7-8f24-c3ae82b867d5.png)

## Steps to Reproduce
Browse  http://0.0.0.0:8080 and click **Import** button.

## Checklist before submitting
- [x] I have confirmed this using the officially supported Docker Compose setup using the `local.yml` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [x] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [x] I provided a **minimal code snippet** or list of steps that reproduces the bug.
- [x] I provided **screenshots** where appropriate
- [x] I filled out all the relevant sections of this template

isms:
This needs to be implemented on the frontend _and_ backend.

adrianjoseph:
I am new to the field of data science (currently completing a Masters Program), but I am very interested in the topic/cause. 
ISMS, not sure if your comment is a statement or a question, but I am curious whether the solution should indeed be bounded by delivery of working code or also is a gui/application also needed?

isms:
> I am curious whether the solution should indeed be bounded by delivery of working code or also is a gui/application also needed?

@adrianjoseph Not sure I understand the question?

To be clear, I mocked out the Vue component (frontend) for looking through directories to find the folder of imagery we want to import as an `ImageSeries` [model](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/interface/backend/images/models.py#L9-L38) with filesystem location and metadata stored in the database (backend).

Neither part is done (100% functional, well tested, etc) so that's why both parts need work.

adrianjoseph:
Ahh

That makes more sense

Appreciate you clarifying - I got my answer.

Let me take a closer look at your code

Thanks



Adrian

jeffreyb92:
[...]

PR #233, opened 2017-11-19
#145 import image series

vessemer:
Import image series implementation. 

## Reference to official issue
This addresses #145 

## Screenshots:
ezgif com-cro

## CLA
- [x] I have signed the CLA; if other committers are in the commit history, they have signed the CLA as well

PR #236, opened 2017-11-19
Implemented Imagery selection screen

Serhiy-Shekhovtsov:
## Description
According to [documentation](http://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#imagery-interface-frontend) on this screen should in particular:

- List available DICOM images on local filesystem.
- Display a preview of an available DICOM image.
- Allow user to select a given image and **start a new “Case.”** This image will be used in the identification step next.

All these features has been implemented in this PR(frontend and backend).
- Now user can see details about currently running case
- Can see list of directories and preview an image
- Can start a new case for selected image

start a new case

Related to #145 and #202

## CLA

### DIFF
--- a/interface/backend/api/views.py
+++ b/interface/backend/api/views.py
@@ -10,6 +10,11 @@ from backend.cases.models import (
     Nodule,
     CaseSerializer
 )
+from backend.cases.factories import (
+    CaseFactory,
+    CandidateFactory,
+    NoduleFactory
+)
 from backend.images.models import ImageSeries
 from django.conf import settings
 from django.core.files.storage import FileSystemStorage
@@ -195,3 +200,30 @@ def nodule_update(request, nodule_id):
     Nodule.objects.filter(pk=nodule_id).update(lung_orientation=enums.LungOrientation[lung_orientation].value)
     return Response(
         {'response': "Lung orientation of nodule {} has been changed to '{}'".format(nodule_id, lung_orientation)})
+
+
+@api_view(['POST'])
+def start_new_case(request):
+    try:
+        uri = json.loads(request.body)['uri']
+    except Exception as e:
+        return Response({'response': "An error occurred: {}".format(e)}, 500)
+
+    # drop existing case(s) with candidates and nodules
+    Case.objects.all().delete()
+
+    # drop imported images
+    ImageSeries.objects.all().delete()
+
+    # import a new image and start a new case
+    new_image, created = ImageSeries.get_or_create(uri)
+    new_case = CaseFactory(series=new_image)
+
+    # TODO: start prediction here
+    # dummy data
+    candidates = CandidateFactory.create_batch(5, case=new_case)
+    NoduleFactory(candidate=candidates[0], case=new_case)
+    NoduleFactory(candidate=candidates[1], case=new_case)
+    NoduleFactory(candidate=candidates[4], case=new_case)
+
+    return Response(CaseSerializer(new_case).data)
--- a/interface/frontend/src/components/open-image/ImageSeries.vue
+++ b/interface/frontend/src/components/open-image/ImageSeries.vue
@@ -1,85 +1,53 @@
 <template>
   <div class="container">
-
-    <div class="row">
-      <div class="col-md-12">
-        <div class="card">
-          <div class="card-header">
-            Open imagery
-          </div>
-          <div class="card-block">
-            <template v-if="availableSeries.length">
-              <ul>
-                <li v-for="series in availableSeries" :key="series.series_instance_uid">
-                  <a href="#" @click="selectSeries(series)">{{ series.series_instance_uid }}</a>
-                  <span v-if="series == selected">&larr;</span>
-                </li>
-              </ul>
-            </template>
-            <template v-else>
-              <p class="card-text">No images imported.</p>
-            </template>
-            <button class="btn btn-warning float-right"
-                    @click="showImport = !showImport"
-            >
-              Import
-            </button>
-          </div>
-        </div>
+    <div class="offset-top">
+      <div class="pt-3">
+        <h1> Open Image </h1>
       </div>
-    </div><!-- /row1 -->
+      <hr/>
+      <div v-if="currentCase">
+        <h4>
+          You have started this case:
+        </h4>
 
-    <div class="row" v-show="showImport">
-      <div class="col-md-12">
-        <div class="card card-outline-warning">
-          <div class="card-header">
-            Import image series
-          </div>
-          <div class="card-block left">
-            <tree-view class="item left" :model="directories"></tree-view>
-            <open-dicom class="right" :view="preview"></open-dicom>
-          </div>
-        </div>
+        <table class="table table-bordered table-condensed">
+          <tr>
+            <td>Patient ID</td>
+            <td>{{ currentCase.series.patient_id }}</td>
+          </tr>
+          <tr>
+            <td>Series Instance ID</td>
+            <td>
+              <small>{{ currentCase.series.series_instance_uid }}</small>
+            </td>
+          </tr>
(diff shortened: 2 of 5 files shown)


interface/backend/api/views.py
```diff
@@ -195,3 +200,32 @@ def nodule_update(request, nodule_id):
...
+
+@api_view(['POST'])
+def start_new_case(request):
+    try:
+        uri = json.loads(request.body)['uri']
+    except Exception as e:
+        return Response({'response': "An error occurred: {}".format(e)}, 500)
+
+    # drop existing case(s) with candidates and nodules
+    all_cases = Case.objects.all()
```
lamby: Hang on, is that right? It will essentially make the entire process global whilst we would surely want it per-client?

interface/backend/api/views.py
```diff
@@ -195,3 +200,32 @@ def nodule_update(request, nodule_id):
...
+    all_cases.delete()
+
+    # drop imported images
+    all_images = ImageSeries.objects.all()
+    all_images.delete()
```
lamby: FYI you can can do ``ImageSeries.objects.all().delete()``

interface/frontend/src/components/open-image/ImageSeries.vue
```diff
@@ -1,85 +1,53 @@
...
-            </template>
-            <button class="btn btn-warning float-right"
-                    @click="showImport = !showImport"
-            >
-              Import
-            </button>
-          </div>
-        </div>
+        <p>
+          You can select an other DICOM image and start a new case
```
louisgv: `an other` -> `another`

interface/frontend/src/components/open-image/ImageSeries.vue
```diff
@@ -97,54 +64,52 @@
...
     },
     mounted: function () {
-      EventBus.$on('dicom-selection', (path) => {
-        this.preview.paths = path
-        console.log(this.preview)
+      EventBus.$on('dicom-selection', (paths) => {
+        this.preview.paths = paths
+      })
+
+      EventBus.$on('start-new-case', (filePath) => {
```
louisgv: I wonder why is this using EventBus? The method to send new case can be invoked from startNewCase directly. If the params filePath is missing, we can simply pass it into this ImageSeries component via the props down in the TreeViews.

I think we should use EventBus sparingly to ensure components are modular and feature-compact. (That is, if I take this component and use it somewhere else, I don't have to worry about emitting an event from the parent component, but I can depend on this component to handle the job)

interface/backend/api/views.py
Serhiy-Shekhovtsov: This a valid concern, but it's a way all our modules implemented right now. We can later add authentication and implement per-client separation.

interface/frontend/src/components/open-image/ImageSeries.vue
Serhiy-Shekhovtsov: @louisgv, I am new to Vue.js so may be wrong but it seems like there is no simple way to implement two way binding between one component and it's ground child. We can emit events and catch+emit them on each middleman, but it doesn't seem nice to me. Also we can use DOM method `this.$el.dispatchEvent` like suggested [here](https://stackoverflow.com/questions/42615445/vuejs-2-0-emit-event-from-grand-child-to-his-grand-parent-component).

louisgv: @Serhiy-Shekhovtsov I think I should have commented on the function down there instead of here :p 

My thought was, is there a way to make the grand-child component feature complete without the need to rely on the grand-parent component? Since if the grand-child needs anything, the grand-parent can simply pass it down. Thus, we can maintain the uni-directional data flow, and will not have to worry about propagating data upstream, and will also ensure the grand-child has all the feature it need to be stand-alone and re-usable.

interface/backend/api/views.py
isms: @lamby Bear in mind the design doc stipulates single standalone computer w/ no auth story to keep everyone focused on the value added parts of the project.

lamby:
@Serhiy-Shekhovtsov @vessemer  I see that this also part of PR #233. How might we best come to a consensus on which to merge..? Or which bits to merge, etc.?

Serhiy-Shekhovtsov:
@lamby my PR fits with requirements and easier for user: you can see images on the local system by default, click Preview and click Start new case.
While PR #233 requires the user to click Import first, click Preview image, then he has to add an image(by clicking Import) and then Create a case.

From my point of view, those steps are redundant and don't match with [requirements](http://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#imagery-interface-api). Of course,  @vessemer can still update his PR, but I can't see much sense in this. If points is an issue - I have no problem with sharing them between us.

vessemer:
@Serhiy-Shekhovtsov, in your PR I'm mainly concern with the `Case` model usage, as it was mentioned in the [review](https://github.com/concept-to-clinic/concept-to-clinic/pull/236#pullrequestreview-77660241). I guess it will be better if we allow to user switch between `Case.objects` without losing them. Also, there should be an opportunity to create more than one `Case.object` for a single `ImageSeries.object` (in case of several observators). My PR #233 aimed to deal with such case while introducing minimum changes into existing interface.
I also guess that it is not a good idea to introduce an `EmitBus` listener in such a general component as `TreeViews`.

lamby:
> I guess it will be better if we allow to user switch between Case.objects without losing them. Also, there should be an opportunity to create more than one Case.object for a single ImageSeries.object (in case of several observators). 

I think this is needed.. or, putting it another way, I could open another issue to do this separate to this.

Serhiy-Shekhovtsov:
Switching between cases is good idea. From my point of view, the following changes could improve the #233:
## No need to show Import button and require user to click it. 
**Show the list of cases** and **the TreeView with files by default**. User should be able to switch the case or start a new case with minimum amount of clicks.
open image

## No need to require user to import image before starting a new case. 
Nor to show the redundant information. On the backend you can use `ImageSeries.get_or_create(uri)` method to load existing ImageSeries for a new Case.

## Accordion like behavior for files TreeView
Because user should see what file he is previewing now. In current version user can unfold multiple folders same time and it's not clear what DICOM is opened in the viewer at the moment. This is the reason I used `EmitBus`. If you can see a better way to implement it - I have nothing against it.

If these changes are OK with you guys, @vessemer can update his PR and I will close my in favor of it. Alternatively, I can merge the list of cases from #233 to my PR.

lamby:
@Serhiy-Shekhovtsov @vessemer

To clarify, it isn't really *switching* between cases it just that the system surely needs to be able to support multiple cases happening independently of each other.. I mean, otherwise we just have introduced a bunch of global variables / state that, well, that hopefully speaks for itself :)

I'd really dearly love for us to get one of #236 or #233 merged and am very happy to award points to both, I'd just need some clarification/confirmation on what exactly what code we should end up.

Did you come to any conclusions on this on another ticket? :)  Let me know and I'll "action" stuff ASAP.

vessemer:
I've implemented the aforementioned features without `EmitBus` usage, by this [commit](https://github.com/concept-to-clinic/concept-to-clinic/pull/233/commits/4c1420fada44e2665b52aaf56a3fba4a24c76745).  
Double the [demonstration](https://github.com/concept-to-clinic/concept-to-clinic/pull/233#issuecomment-346517671) here:
peek 2017-11-23 02-38

lamby:
Loving the GIF  screenshots btw - keep them coming! So, I'm seeing some conflicts here (and on #233 FYI).

@Serhiy-Shekhovtsov Any input from thee? :)

Serhiy-Shekhovtsov:
Guys, I am closing this PR in favor of #233. Will post my thoughts there.

Issue #204, opened 2017-11-02
Prevent progress to next stage of analysis until "completed" the current step


lamby:
## Expected Behavior
The interface should prevent you from moving to the next stage of the image analysis until you have completed the current step, whatever that means for each step.

## Current Behavior
You can proceed to, for example, "Annotate and segment" wthout importing any images.

## Checklist before submitting
- [ ] I have confirmed this using the officially supported Docker Compose setup using the `local.yml` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [ ] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [ ] I provided a **minimal code snippet** or list of steps that reproduces the bug.
- [ ] I provided **screenshots** where appropriate
- [ ] I filled out all the relevant sections of this template

louisgv:
Will be thinking about a good practice solution for this.

louisgv:
If nobody has working on this yet, I will begin tomorrow.

louisgv:
So I think since each tab are controlled by vue-router, the solution to this would be to add a gate layer to each route and check for condition. 

https://router.vuejs.org/en/advanced/navigation-guards.html

Will be going through the doc and implement a simple gate layer for each route.

PR #239, opened 2017-11-20
Router Guard + Refactor Front-end Test Suite

louisgv:
Per #204, #245  #246  I initialized a vue-router guard as well as refactor and make front-end unit and e2e test functional.

## Description

The front-end unit and e2e test are now functional and can be run off a docker environment. The instruction to run the tests are documented and the test commands are also added to the global test script.

The E2E test is now using an external selenium instance via selenium-grid docker container.

The unit test is now using AVA test engine instead of Karma for its better async/test API as well as concurrency support.

This effectively allow us to, (since we are implementing all of our ui layer under the component), use the global guard to make sure subsequent route has the required data from the previous route, using a very primitive `Store` singleton that I derived. 

It's very flexible in term of refactor into using either a Redis cache or localstorage, and can be imported into any other component. This way we do not have to refactor the data layer of each component to propagate the data back to its parent, but can simply import `Store` as follow:

```
import Store from '../store'
```

In any of our component, then simply call the static methods defined in the `Store` class to update the dependency data. (right now I simply test with `this.$route.path` )

#204 
#245 
#246 

## Motivation and Context

+ This will make E2E and unit test functional again, allowing future work to be programmatically verified.
+ This effectively ensure subsequent view has all the required data needed before render. 
+ Without a gate, view without dependency data will be empty.

> NOTE: When we need more complicated state handling, we can look into a more robust solution like https://vuex.vuejs.org, tho their code is quite bloated (for good reason, e.g vue dev-tool binding, error handling etc...). The current `Store` implementation can be refactored to use vuex easily if we ever need it. For now, I would avoid introducing new dependency.

## How Has This Been Tested?

I implemented Unit test for Store and E2E test for going to unauthorized route.

ezgif com-gif-maker

## CLA


WGierke:
Cool stuff! Is it possible to introduce tests for the front-end as well?

louisgv:
@WGierke For sure, I can take a whack at some unit test for store as well as a simple E2E test of trying to go to future route without prerequisite.

lamby:
:+1: for the tests - will hold off merging this until you've given it a stab :)

louisgv:
I will be adding selenium to our docker setup for E2E headless testing:
https://github.com/mycargus/nightwatch-docker-grid

Referencing the repo for future documentation/recognition

lamby:
How are we looking @louisgv? (apart from the file conflict..)  :)

louisgv:
@lamby E2E test is done. I'm very close to get unit test done. I'm revamping the tooling as well as adding a sample unit test.

Also going to improve the E2E test a little bit to make it more concise. I expect this PR to be good to review by tomorrow.

louisgv:
@WGierke @lamby Unit test and E2E test for Frontend are fixed and are now running. The test case for the route and store was also implemented.

louisgv:
@WGierke @lamby can you manually trigger Travis to re-build. It is failing on the python backend side for some reason.

WGierke:
I'm not able to do that but you can do a `git commit --amend` and `git push -f` to manually restart the Travis test. However, give me a second to file an issue for that and (hopefully) to directly propose a patch :)

lamby:
> Took me the entire day since I had to re-build docker over and over again to retry different test setup to make sure they run :package:

I'm so sorry. :)

.. and yes, what @WGierke said re. re-running the tests.. :)

louisgv:
@WGierke @lamby Travis passed, BLESS :clap:

lamby:
Thanks @louisgv. Two quick final things:

* I think you'll need to rebase against the `master` branch after these recent changes.
* Just confirming that you believe this addresses all three of #204, #245 & #246?

louisgv:
@lamby Yes, this should address all three issue. Detail are noted in the PR description. I also rebased with master as requested.

lamby:
Thanks!

Issue #229, opened 2017-11-17
Algorithms don't work across both dicom and luna scans


reubano:
## Overview

Currently, the classification algorithm is tested with a luna scan, and the segmentation and identification algorithms are tested with dicom scans. This indicates that there may be some incompatibilities between dicom and luna scans that make them not work with certain algorithms. I have confirmed this by adding new tests for the complementary scans and produced the following results: 

algorithm|dicom|luna
---------|-------|----
classification|FAILS|PASSES
segmentation|PASSES|FAILS
identification|*crashes*|*crashes*

## Expected Behavior

All algorithm tests should pass

## Technical details

Below are the specific tracebacks:

`test_classification.py`

```bash
(alcf) ⚡ docker-compose -f local.yml run prediction pytest -vrsk src/tests/test_classification.py
========================================== test session starts ==========================================
platform darwin -- Python 3.6.2, pytest-3.1.3, py-1.4.34, pluggy-0.4.0 -- /opt/local/bin/python
cachedir: .cache
rootdir: /Users/reubano/Documents/Projects/alcf/prediction, inifile:
collected 3 items

src/tests/test_classification.py::test_classify_predict_load PASSED
src/tests/test_classification.py::test_classify_dicom FAILED
src/tests/test_classification.py::test_classify_luna PASSED

=============================================== FAILURES ================================================
__________________________________________ test_classify_dicom __________________________________________

dicom_paths = ['/Users/reubano/Documents/Projects/alcf/images_full/LIDC-IDRI-0002/1.3.6.1.4.1.14519.5.2.1.6279.6001.4901573811602007...14519.5.2.1.6279.6001.298806137288633453246975630178/1.3.6.1.4.1.14519.5.2.1.6279.6001.179049373636438705059720603192']
nodule_locations = [{'x': 50, 'y': 50, 'z': 21}]
model_path = '/Users/reubano/Documents/Projects/alcf/prediction/src/algorithms/classify/assets/gtr123_model.ckpt'

    def test_classify_dicom(dicom_paths, nodule_locations, model_path):
>       predicted = trained_model.predict(dicom_paths, nodule_locations, model_path)

src/tests/test_classification.py:9:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[...]

Issue #195, opened 2017-10-30
Stop slow tests after a set timeout period, instead of skipping them


reubano:
## Overview

As made evident by #191, skipping slow tests potentially hides bugs. These test should instead be stopped after a set period of time, e.g. ~15s, to ensure they don't cause errors.

## Expected Behavior

The identification tests should fail on travis even when `RUN_SLOW_TESTS` isn't set to `true`.

## Technical details

The desired behavior should be able to be accomplished by combining the [pytest-timeout](https://pypi.python.org/pypi/pytest-timeout#usage) package with the [`pytest.mark.xfail` raises-parameter](https://docs.pytest.org/en/latest/skipping.html#raises-parameter). Setting the environment variable `RUN_SLOW_TESTS=true` should cause the timeout parameters to be ignored.

## Acceptance criteria

 - [ ] Running on travis with `RUN_SLOW_TESTS=false` (the default) should show the `src/tests/test_endpoints.py::test_identify` function as failed. 
 - [ ] Running on travis with `RUN_SLOW_TESTS=false` (the default) should show the `src/tests/test_segmentation.py::test_lung_segmentation` function as passed. 
 - [ ] Running on travis with `RUN_SLOW_TESTS=true` should show the cause travis to timeout the build. 

*NOTE: All PRs must follow the standard PR checklist.*

WGierke:
`The identification tests should fail on travis even when RUN_SLOW_TESTS isn't set to true.`
Assuming that an error occurs in the seconds before the timeout, right? Otherwise the test passes because the timeout is not seen as an error?

reubano:
@WGierke yes, exactly.

WGierke:
If I understand [this](https://stackoverflow.com/questions/46766899/pytest-timeout-fail-test-instead-killing-whole-test-run) correctly, it doesn't appear to be so easy to catch the case when pytest-timeout stops a method because the SIGALRM signal does not produce an exception. I was able to implement the remaining logic but currently I'm struggling with somehow catching that timeout exception:
`git checkout bcb743766386794bf373bd14972cb3b05c60d727`  
`git apply patch.txt`
[patch.txt](https://github.com/concept-to-clinic/concept-to-clinic/files/1485956/patch.txt)
Does someone have an idea how to still let tests pass even when they timeout?

WGierke:
Nvm, I think I got it :)

Issue #191, opened 2017-10-26
Nodule identification doesn't work


reubano:
## Overview

The nodule Identification `predict` function errors

## Expected Behavior

The `predict` function should return a value without erroring

## Technical details

`cd prediction`

```python
>>> from src.algorithms.identify.trained_model import predict
>>> 
>>> dicom_path = '../images/LIDC-IDRI-0001/1.3.6.1.4.1.14519.5.2.1.6279.6001.298806137288633453246975630178/1.3.6.1.4.1.14519.5.2.1.6279.6001.179049373636438705059720603192'
>>> predict(dicom_path)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-9-71ee08c0f3e1> in <module>()
----> 1 predict(dicom_path)

~/Documents/Projects/alcf/prediction/src/algorithms/identify/trained_model.py in predict(dicom_path)
     48     reader.SetFileNames(reader.GetGDCMSeriesFileNames(dicom_path))
     49     image = reader.Execute()
---> 50     result = gtr123_model.predict(image)
     51     return result
     52     # if dicom_path[-1] != '/':

~/Documents/Projects/alcf/prediction/src/algorithms/identify/src/gtr123_model.py in predict(ct_path, model_path)
    459 
    460     """
--> 461     ct_array, meta = load_ct.load_ct(ct_path)
    462     meta = load_ct.MetaImage(meta)
    463     spacing = np.array(meta.spacing)

~/Documents/Projects/alcf/prediction/src/preprocess/load_ct.py in load_ct(path, voxel)
     94             of a MetaImage file or DICOM-series in its original format.
     95     """
---> 96     dicom_pattern = os.path.join(path, '*.dcm')
     97     dicom_pattern = glob(dicom_pattern)
     98     mhd_pattern = os.path.join(path, '*.mhd')

/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/posixpath.py in join(a, *p)
     76     will be discarded.  An empty last part will result in a path that
     77     ends with a separator."""
---> 78     a = os.fspath(a)
     79     sep = _get_sep(a)
     80     path = a

TypeError: expected str, bytes or os.PathLike object, not Image
```

This error first appears in commit [#c448e44](https://github.com/concept-to-clinic/concept-to-clinic/commit/c448e44c65eb49a12baf1731f273f118d34f0bbf)

[...]

PR #243, opened 2017-11-21
#195 Stop slow tests after set period of time

WGierke:
We don't want to completely skip slow tests since this can hide bugs (e.g. #191). Instead, we want to start the slow tests but skip them after a specified period of time.

## Description
If the environment variable RUN_SLOW_TESTS is set to True, all tests should be run. Otherwise tests that exceed the time specified by the variable TESTS_TIMEOUT (default is 15 seconds) are skipped without causing an error.

This addresses #195 

## Motivation and Context
It was necessary to create a new [hook](https://github.com/concept-to-clinic/concept-to-clinic/compare/master...WGierke:195_stop_slow_tests?expand=1#diff-507495becfd1e874bad6e0b1cf4437b2R115) that checks whether a test should timeout, that kills the test when exceeding the timeout and then returns that the test is not failing (if the timeout is the only reason for the test exit). It wasn't possible to just use pytest-timeout and the xfail marker since pytest-timeout does not raise an exception when killing a test which could be caught by the xfail marker.

## How Has This Been Tested?
I added a [test](https://github.com/concept-to-clinic/concept-to-clinic/compare/master...WGierke:195_stop_slow_tests?expand=1#diff-ed63136572f32bc5f7500b864f9bdfa9R60) that should always timeout (if we want slow tests to timeout), otherwise it raises an uncaught exception. The tests that are failing are expected to fail (as stated in the issue and #229 ) namely 
`src/tests/test_endpoints.py::test_identify`, `src/tests/test_identification.py::test_identify_nodules_001` and `src/tests/test_identification.py::test_identify_nodules_003`.

## CLA

### DIFF
--- a/prediction/src/tests/conftest.py
+++ b/prediction/src/tests/conftest.py
@@ -1,9 +1,11 @@
-from os import path
+import signal
 from glob import glob
+from os import path
 
 import pytest
-
 from config import Config
+
+from . import get_timeout
 
 
 @pytest.fixture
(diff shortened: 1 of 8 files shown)


tests/test_docker.sh
```diff
@@ -7,7 +7,7 @@
 set -ex
 
 # run the model service's tests
-docker-compose -f local.yml run prediction pytest -rs
+docker-compose -f local.yml run prediction pytest -rsx
```
WGierke: `x` gives a report on xfails e.g.
```
XFAIL src/tests/test_segmentation.py::test_lung_segmentation
  Test was stopped after timeout
XFAIL src/tests/test_segmentation.py::test_stop_timeout
  Test was stopped after timeout
```

lamby:
Neat :)

WGierke:
Was that a LGTM? :grin:

prediction/src/tests/conftest.py
```diff
@@ -81,3 +83,54 @@ def model_path(scope='session'):
...
+    timeout = get_timeout()
+    found_timeout_marker = False
+
+    for k in item.keywords:
+        found_timeout_marker |= (k == "stop_timeout")
+
+    if timeout == 0 or not found_timeout_marker:
+        # All slow tests should be run or the test does not have a stop_timeout marker
+        yield
+        return
```
reubano: Why both `yield` *and* `return`?

reubano:
From travis

```
$ flake8 prediction
prediction/src/tests/test_endpoints.py:19:1: E303 too many blank lines (3)
The command "flake8 prediction" exited with 1.

$ pycodestyle prediction
prediction/src/tests/test_endpoints.py:19:1: E303 too many blank lines (3)
The command "pycodestyle prediction" exited with 1.
```

reubano:
> The tests that are failing are expected to fail (as stated in the issue and #229) namely

Yep I noticed this. Is it possible to set a timeout lower than then memory error so that we won't have to wait until #229 is fixed?

prediction/src/tests/conftest.py
WGierke: The hook needs to yield exactly once, otherwise there'll be an error. Without the `return` it would yield twice, without the `yield` it wouldn't yield at all. Maybe there is another kind of hook one could use but using this template from SO worked pretty nice so far...

prediction/src/tests/test_segmentation.py
```diff
@@ -52,3 +54,11 @@ def test_lung_segmentation(dicom_paths):
             z_mask = int(abs(min_z) - abs(centroid_z))
             mask_value = patient_mask[z_mask, x_mask, y_mask]
             assert mask_value == 255
+
+
+@pytest.mark.stop_timeout
+def test_stop_timeout():
+    timeout = get_timeout()
+    if timeout > 0:
+        time.sleep(get_timeout() + 1)
```
reubano: `time.sleep(timeout + 1)`?

prediction/src/tests/conftest.py
reubano: gotcha!

reubano: Mind adding that as a comment in the code? I'm sure future you and others will appreciate it :).

reubano:
After fixing the minor conflict, it looks good to go now. Mind squashing?

Issue #247, opened 2017-11-23
Race Condition with Timeout Tests


WGierke:
As discussed in #239 , apparently #243 introduced a race condition that involves tests that should timeout. Sometimes, a `TimeoutExit` is thrown (which is correctly caught) and sometimes a `Termination` exception is thrown which is currently now caught and makes the tests fail (e.g. [here](https://travis-ci.org/concept-to-clinic/concept-to-clinic/builds/306144770?utm_source=github_status&utm_medium=notification))

## Expected Behavior
A test that creates a timeout should be caught by `xfail` and thus not fail.

## Possible Solution
Only throw `TimeoutExit` exceptions since there is no reason to throw two kinds of exceptions when a test should be killed.

## Checklist before submitting
- [X] I have confirmed this using the officially supported Docker Compose setup using the `local.yml` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [X] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [X] I filled out all the relevant sections of this template

PR #248, opened 2017-11-23
#247 Only use TimeoutExit for timeouts

WGierke:
This PR fixed the case that a `Termination` exception can be thrown in tests that should timeout.

This addresses #247 .

## CLA


lamby:
Cheers for this :)

PR #249, opened 2017-11-24
implemented area select vue component for Segmentation page (#150)

Serhiy-Shekhovtsov:
Implemented Vue.js component for area selection based on [this project](https://github.com/neshte/jquery-canvas-area-draw). Did following updates:

- converted to Vue component
- get rid of jQuery dependency
- added ability to move entire area when `Shift` is pressed

vue-area-select

Can be used as simple as this:
`<area-select @selection-changed="areaSelectChange" v-if="showAreaSelect" :selectedArea="currentArea"></area-select>`

#150

## License
Since I used [this project](https://github.com/neshte/jquery-canvas-area-draw) as a starting point, should I include it's license in the code?

## CLA

### DIFF
--- a/interface/frontend/src/components/open-image/AreaSelect.vue
+++ b/interface/frontend/src/components/open-image/AreaSelect.vue
@@ -1,0 +1,273 @@
+<template>
+  <canvas ref="canvas" @contextmenu="onRightClick" @mousedown="onMouseDown" @mousemove="onMouseMove"
+          @mouseup="onMouseUp" :width="canvasWidth" :height="canvasHeight" :style="canvasStyle">
+  </canvas>
+</template>
+
+<script>
+  // how close near the point should user click to grasp it
+  const graspSensitivity = 6
+
+  // selected area style
+  const strokePolygonRgb = 'rgb(255,255,255)'
+  const fillPolygonRgba = 'rgba(255, 255, 90, 0.2)'
+
+  // point style
+  const strokeCoordRgb = 'rgb(255, 255, 90)'
+  const fillCoordRgb = 'rgb(255, 255, 90)'
+
+  export default {
+    components: {},
+    props: ['width', 'height', 'selectedArea'],
+    data () {
+      return {
+        areaCoordinates: this.selectedArea || [],
+        draggingPointIndex: -1,
+        lastMouseLocation: null,
+        draggingEntireArea: false,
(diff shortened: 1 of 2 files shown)


interface/frontend/src/components/open-image/AreaSelect.vue
```diff
@@ -0,0 +1,271 @@
...
+        if (e.preventDefault) e.preventDefault()
+      },
+      getMouseLocation (e) {
+        const canvasOffset = this.getElementOffset(this.canvas)
+        const x = (e.pageX - canvasOffset.left)
+        const y = (e.pageY - canvasOffset.top)
+
+        return {x, y}
+      },
+      pointToLineLength (x, y, x0, y0, x1, y1, o) {
```
lamby: Dat math. Was this code, err, borrowed from somewhere? If so, we should really credit it :)

interface/frontend/src/components/open-image/OpenDICOM.vue
```diff
@@ -124,6 +127,9 @@
       }
     },
     methods: {
+      areaSelectChange (newCoords) {
+        console.log('areaSelectChanged', JSON.stringify(newCoords))
```
lamby: Out of interest, do we want to be logging directly to `console.log` here? :)

Serhiy-Shekhovtsov: The component I have implemented will be used on Segmentation page, but currently it [is disabled here](https://github.com/concept-to-clinic/concept-to-clinic/pull/249/files#diff-70d179422138a3042c2135a282065626R64), so there will be no logs. And once someone will start working on that page, he will enable the component at that line and handle user input at this line.

interface/frontend/src/components/open-image/AreaSelect.vue
Serhiy-Shekhovtsov: For `pointToLineLength` it was used by [original repo](https://github.com/neshte/jquery-canvas-area-draw) without mentioning of the author. But I have found, that this method comes [from here](http://jsfromhell.com/math/dot-line-length). Will include the reference.
My other question remains open - should we add the license and reference to [original repo](https://github.com/neshte/jquery-canvas-area-draw)?

lamby:
Thanks!

Issue #244, opened 2017-11-21
Update that version number in the Navbar


pjbull:
## Expected Behavior
Navbar has short git hash that matches the current version of the code. This will be especially valuable on the [demo site](http://demo.concepttoclinic.drivendata.org).

## Current Behavior
Nav bar displays the string `a0b1c2d` all the time.

<img width="1000" alt="screen shot 2017-11-21 at 11 46 32 am" src="https://user-images.githubusercontent.com/1799186/33101365-eba44e06-cecc-11e7-9b3c-9d1370eecbcb.png">

## Steps to Reproduce
1. Run the application UI

## Possible Implementation
 - Get short git hash in the [backend Dockerfile](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/compose/interface/Dockerfile-dev-api)
 - Export to `GIT_HASH_VERSION` env variable in Dockerfile (so that it is evaluated _at container build time_)
 - Get that env variable within the backend settings
 - Return that setting value through an endpoint (actually, maybe through all of them?)
 - Call that [endpoint in the frontend navbar](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/interface/frontend/src/components/AppNav.vue) to get the right data

lamby:
Merged in https://github.com/concept-to-clinic/concept-to-clinic/commit/615261c5cebf040dc971fd212b161c7fbd1055d1 :)

WGierke:
@lamby can this issue be closed then?

PR #253, opened 2017-11-27
#244 Update the version number in the navbar

musale:
Updates the short git hash that matches the current version of the code in the navbar

## Description
This allows the navbar to have the short version of the current commit hash from HEAD.

Fixes issue #244 

## Motivation and Context
Get the hash of the latest commit as the display instead of a static value.

## How Has This Been Tested?
- Run `docker-compose -f local.yml build`
- Run `docker-compose -f local.yml up`
- View the application UI on [http://localhost:8080/#/](http://localhost:8080/#/)
- Open another terminal in the root of your project repo and run `git rev-parse --short HEAD`
- The hash on the top right side of the navbar should be the same to the value of the previous step

## CLA

### DIFF
--- a/local.yml
+++ b/local.yml
@@ -55,7 +55,7 @@ services:
       - ./interface/frontend/dist:/app/dist:cached
       - ./interface/frontend/static:/app/static:cached
       - ./interface/frontend/test:/app/test:cached
-      - ./.git/logs/HEAD:/HEAD:cached
+      - ./.git/logs/HEAD:/app/HEAD:cached
     links:
       - interface
     ports:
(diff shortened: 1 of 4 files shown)


local.yml
```diff
@@ -55,7 +55,7 @@ services:
       - ./interface/frontend/dist:/app/dist:cached
       - ./interface/frontend/static:/app/static:cached
       - ./interface/frontend/test:/app/test:cached
-      - ./.git/logs/HEAD:/HEAD:cached
+      - ./.git:/.git:cached
```
lamby: Hm, the .git subdir is big - is there not a smaller subset we could use instead? Or use `logs/HEAD`?

musale: Actually we could use the `logs/HEAD` file and read the contents with some tweaking. Let me update that.

musale:
@lamby it's updated now 😃

reubano:
The travis build erred, so I restarted it.

lamby:
Thakns"

PR #250, opened 2017-11-25
Detect and select: attaching nodule centroid marker to current slice

Serhiy-Shekhovtsov:
Implemented slice switching for candidate list, and a way for user to see when slice differs and attach the marker to current slice.

## Description
Following features has been implemented:
- switching the viewer to candidate's slice when selecting a candidate
- showing to user that marker is on an other slice
- clicking on a marker attaches it to current slice

On back-end I have added a method for loading candidates with cases and series. Series object will also have list of DICOM files in the folder saved in `uri` property.

## Reference to official issue
It's an improvement of #148

## Screenshot:
detect-and-select-z
[full size image](http://g.recordit.co/6AwCrut5WW.gif)

## CLA
- [x] I have signed the CLA; if other committers are in the commit history, they have signed the CLA as well

PR #254, opened 2017-11-28
Detect and select: mark\dismiss candidates

Serhiy-Shekhovtsov:
Implemented functionality for reviewing nodule candidates. Now user can set whether the nodule was marked for further analysis or not.

**NOTE:** this PR depends on #250 and has only 3 commits on top of it.

It's an improvement of #148

## Screenshot:
detect-and-select-review
[full size image](http://g.recordit.co/MlvpaNE9cB.gif)

## CLA


reubano:
Works like a charm! Rebasing should sort out the errors :)

interface/backend/cases/tests.py
```diff
@@ -118,3 +118,21 @@ def test_update_nodule_lung_orientation(self):
...
+        self.assertEquals(candidate.review_result, enums.CandidateReviewResult.MARKED)
+
+    def test_candidates_dismiss(self):
+        candidate = CandidateFactory()
+        url = reverse('review-candidate', kwargs={'candidate_id': candidate.id})
+        self.client.post(url, json.dumps({'review_result': enums.CandidateReviewResult.DISMISSED}), 'application/json')
+
+        candidate.refresh_from_db()
+
+        self.assertEquals(candidate.review_result, enums.CandidateReviewResult.DISMISSED)
```
reubano: Ohh, I see the tests were just moved over here :) Is this a better location for them?

Serhiy-Shekhovtsov: I just saw `test_update_nodule_lung_orientation` located here and used it as an example. But, ideally, we would test both aspects - the server response and result state in DB. The thing is - I am not sure what is the test coverage policy in our project at this stage.

reubano: Yes, I do now see the tests are a bit different. What issues did you run into that you decided to remove the above ones?

Serhiy-Shekhovtsov: The API has been changed. It used to return only a message about marking\dismissing, I have changed it to return the updated candidate. It's not like I was not able to fix the test because of some issue. Just found an example with `test_update_nodule_lung_orientation` and added the same type of test instead of old one.

interface/backend/api/tests.py
```diff
@@ -41,19 +41,3 @@ def test_images_metadata_view(self):
...
-        self.assertEqual(response_dict["response"], "Candidate {} was marked".format(candidate.id))
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-
-    def test_candidates_dismiss(self):
-        candidate = CandidateFactory()
-        url = reverse('candidate-dismiss', kwargs={'candidate_id': candidate.id})
-        response = self.client.get(url)
-        response_dict = response.json()
-        self.assertEqual(response_dict["response"], "Candidate {} was dismissed".format(candidate.id))
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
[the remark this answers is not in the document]
reubano: > I have changed it to return the updated candidate.

So how about something like this...

```python
def test_candidates_mark(self):
    candidate = CandidateFactory()
    url = reverse('review-candidate', kwargs={'candidate_id': candidate.id})
    response = self.client.post(url, json.dumps({'review_result': enums.CandidateReviewResult.MARKED}), 'application/json')
    response_dict = response.json()
    self.assertEquals(response_dict, 'SOMETHING')
    self.assertEqual(response.status_code, status.HTTP_200_OK)
```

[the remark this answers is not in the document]
reubano: > The thing is - I am not sure what is the test coverage policy in our project at this stage.

I'd say for changes, just try to make sure the previous tests still pass (making modifications to the test cases where necessary).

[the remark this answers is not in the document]
Serhiy-Shekhovtsov: Yes, something like that should do. Also we'll need to update the property on local candidate and use `CandidateSerializer(candidate, context={'request': None})` to make `SOMETHING` :)
So if we need these tests - I can add them.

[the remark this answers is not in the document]
reubano: Sounds like a plan :). Doesn't need to be anything complex. Just something that lets you verify from the response that everything worked as expected. Also, since it's `POST` (and no longer `GET`) am I correct in assuming the status code won't be `200` anymore?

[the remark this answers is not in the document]
Serhiy-Shekhovtsov: I think it's still 200 in case of success :)

Serhiy-Shekhovtsov:
I still have to add API test for this feature. Will make a new PR for that.

interface/backend/api/views.py
```diff
@@ -179,24 +179,35 @@ def candidates_info(request):
...
     return Response(serialized_candidates)
 
 
-@api_view(['GET'])
-def candidate_mark(request, candidate_id):
-    return Response({'response': "Candidate {} was marked".format(candidate_id)})
+@api_view(['POST'])
+def review_candidate(request, candidate_id):
+    try:
+        review_result = json.loads(request.body)['review_result']
```
isms: @Serhiy-Shekhovtsov For the record, `request.data` does this for you (http://www.django-rest-framework.org/api-guide/requests/#data)

reubano:
> I still have to add API test for this feature. Will make a new PR for that.
 
@Serhiy-Shekhovtsov if you don't have time to do this fairly soon, can you create an issue for it?

isms:
@reubano it's already pretty much taken care of: https://github.com/concept-to-clinic/concept-to-clinic/blob/master/interface/backend/cases/tests.py#L105-L119

reubano:
@isms from my understanding, those are new DB state tests. The [previous server response tests](https://github.com/concept-to-clinic/concept-to-clinic/pull/254/files#diff-8bd18167425784c3d5284ee13a570db5L44) were removed. Are you saying these are actually one in the same?

Serhiy-Shekhovtsov:
Guys, test are a little bit different. Old tests were checking the API output while new one is testing outcome(the DB result).

interface/backend/api/views.py
Serhiy-Shekhovtsov: @isms thanks for sharing. I can see you've updated the code to support **partial_update**.
Btw, the `append_files_to_candidate` is rather a workaround. We should load all files for selected case only once at case start.

PR #261, opened 2017-12-06
Refactor DICOM Processing and ImageFile Storage and Serializers

pjbull:
The goal of this is to significantly simplify a lot of how the backend works. This is done in tandem with some frontend changes that are currently WIP by @isms.

The major changes are:
 - Create an `ImageFile` model that manages loading DICOM images, storing them in the database, knowing where these files are on disk, and being able to do all DICOM work in one place. Also, be explicit in the DICOM properties that we parse.
 - Update Serializers so (1) they all live in the same place, (2) the `Case` serializer can track all of the info we need to care about in a single nested object.
 - Remove `NoduleFeatures` which can just live on the `Nodule` model
 - Rename `MetadataApi` to `Preview` for clarity
 - More robust testing of the API

Left todo:
 - [Remove old APIs in favor of true REST API](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/interface/backend/api/views.py#L190-L236)
 - Process MHD + Raw files instead of DICOM (e.g., LUNA)


interface/frontend/src/components/detect-and-select/CandidateList.vue
```diff
@@ -133,10 +133,13 @@
                 // extending result data with technical properties
                 candidate._saving = false
 
-                let series = candidate.case.series
-                candidate._filesPaths = series.files.map((fileName) => {
-                  return series.uri + '/' + fileName
-                })
+                // REMOVED UNITL STORE IS IMPLEMENTED AND WE CAN GET THESE FROM
```
lamby: ♥

interface/backend/images/models.py
```diff
@@ -1,10 +1,147 @@
...
+        logger = logging.getLogger(__name__)
+        for k, v in self.load_dicom_data_from_disk(self.path)['metadata'].items():
+            # check that the field is a property on the object
+            if not hasattr(self.__class__, k):
+                logger.warning(f"Property '{k}' not a field on ImageFile, discarding value '{v}'.")
+
+            else:
+                self.__setattr__(k, v)
+
+    @classmethod
```
lamby: Tempted to put these classmethods on a model manager instead? It ends up cleaner IME.

interface/backend/images/models.py
```diff
@@ -1,10 +1,147 @@
...
 from django.db import models
+from django.core.exceptions import PermissionDenied
 from django.utils._os import safe_join
 
 
+class ImageFile(models.Model):
+    """ Model for an individual file for an image--that is, a single DICOM slice
+        as it is represented on disk.
+    """
+    class Meta:
```
lamby: FYI "official" Django model ordering puts this after fields

interface/backend/images/models.py
```diff
@@ -1,10 +1,147 @@
...
+        'RescaleSlope': lambda x: {'rescale_slope': int(x)},
+        'Rows': 'rows',
+        'Columns': 'columns',
+        'PixelSpacing': lambda x: {'pixel_spacing_col': float(x[0]), 'pixel_spacing_row': float(x[1])},
+    }
+
+    series = models.ForeignKey('ImageSeries', related_name='images', on_delete=models.CASCADE)
+
+    slice_location = models.FloatField(null=True)
+
```
lamby: Something about all this spacing irks me :)

compose/interface/Dockerfile-dev-api
```diff
@@ -9,6 +9,8 @@ COPY ./compose/interface/start-dev.sh /start-dev.sh
 RUN sed -i 's/\r//' /start-dev.sh
 RUN chmod +x /start-dev.sh
 
+EXPOSE 8000
```
lamby: Can you add this commit's comment as an inline comment?

interface/backend/images/models.py
isms: This will be left as an exercise for the reader :smile_cat:

interface/backend/images/models.py
```diff
@@ -1,18 +1,147 @@
...
+        logger = logging.getLogger(__name__)
+
+        # only allow a key to be read from a single DICOM property
+        # and warn on attempted overwrite
+        def _update_once(d1, d2):
+            for k, v in d2.items():
+                if k in d1:
+                    logger.warning(f"Tried to set '{k}' more than once, ignoring value '{v}'.")
+                else:
+                    d1[k] = v
```
reubano: perhaps move this func up a level so since it doesn't depend on anything provided by the parent func.

interface/backend/images/models.py
```diff
@@ -1,18 +1,147 @@
...
+
+        metadata = dict()
+        for dicom_prop, model_attribute in cls.DICOM_PROPERTIES.items():
+                if callable(model_attribute):
+                    try:
+                        processed_values = model_attribute(dcm_data.get(dicom_prop, None))
+                        _update_once(metadata, processed_values)
+                    except:  # noqa
+                        # bare except is normally unforgivable but we _really_ don't care what went wrong here --
```
reubano: 😆

interface/backend/images/models.py
```diff
@@ -1,18 +1,147 @@
...
+        for k, v in self.load_dicom_data_from_disk(self.path)['metadata'].items():
+            # check that the field is a property on the object
+            if not hasattr(self.__class__, k):
+                logger.warning(f"Property '{k}' not a field on ImageFile, discarding value '{v}'.")
+
+            else:
+                self.__setattr__(k, v)
+
+    @classmethod
+    def load_dicom_data_from_disk(cls, filepath, parse_metadata=True, encode_image_data=False):
```
reubano: Maybe a just a matter of style, but based on class methods i've seen how about simplifying to just `from_disk`? This assumes `ImageFile` is only for DICOM files of course, otherwise something like `dicom_from_disk` would work.

interface/backend/images/models.py
```diff
@@ -1,18 +1,147 @@
...
+        return self.load_dicom_data_from_disk(self.path, parse_metadata=False, encode_image_data=True)['image']
+
+    @staticmethod
+    def _pixel_data2str(buf):
+        _min, _max = buf.min(), buf.max()
+        buf = 254 * (np.array(buf, dtype=np.float) - _min) / (_max - _min) + 1
+        return buf.astype(np.uint16)
+
+    @classmethod
+    def _dicom_to_base64(cls, ds):
```
reubano: similar to above, how would `to_base64` work?

interface/backend/images/tests.py
```diff
@@ -14,13 +20,163 @@ def test_get_create_image_series(self):
...
+        })
+
+        # expected behavior is to log a warning
+        with self.assertLogs(level='WARN') as cm:
+            ImageFile(path=path, series=image_series).save()
+
+        self.assertEqual(len(cm.output), 1)
+        self.assertIn("Tried to set 'rows' more than once", cm.output[0])
+
+        ImageFile.DICOM_PROPERTIES.pop('DoesNotExist2')
```
reubano: Can `DICOM_PROPERTIES` updates be wrapped in setup/cleanup steps so that the manual resetting won't be needed?

interface/backend/api/views.py
```diff
@@ -217,20 +219,11 @@ def update_candidate_location(request, candidate_id):
...
-    return Response(CaseSerializer(case).data)
-
-
-def append_files_to_candidate(candidate):
-    series = candidate['case']['series']
-
-    if 'files' not in series:
-        # using `glob1` as it returns filenames without a directory path
-        series['files'] = glob.glob1(series['uri'], '*.dcm')
+    return Response(serializers.CaseSerializer(case, context={'request': None}).data)
```
reubano: Perhaps set a high level `context = {'request': None}` var since it is used so much?

PR #263, opened 2017-12-06
Add Vuex state for managing the case beginning-to-end

isms:
Remove view guards (at least for right now) - the plan is just not to show UI for elements based on what is in the global state (e.g. no case yet ⇒ don't show detect and select, no marked candidates yet ⇒ don't show annotate and segment, etc).


pjbull:
LGTM, go go vuex store!

reubano:
guess this addresses #264?

pjbull:
@reubano @lamby More fun to review. This branch also still has a lot to be turned into issues. The goal here was to get most of the API calls into `state.js` except for on "Open Image", which needs to do a little extra work.

TODOs, which we'll probably turn some of these into issues.
 - Refactor the OpenDicom viewer to be more succinct vue code
 - Pull out old API style endpoints
 - Create in-memory model objects and serialize those rather than calling the DICOM loading methods on the model
 - Error handling

reubano:
@pjbull looking now. btw, I commented on #261, but since it was already merged, not sure if it sent you guys the notification. just fyi..

interface/frontend/src/components/common/OpenDICOM.vue
```diff
@@ -154,6 +156,23 @@
...
+          if (this.stack.currentImageIdIndex < 0) this.stack.currentImageIdIndex = 0
+        }
+
+        this.stack.imageIds = this.view.paths.map((path) => {
+          return this.view.type + this.view.prefixCS + path
+        }, this)
+        cornerstone.disable(element)
+        this.initCS(element)
+      },
       str2pixelData (str) {
```
reubano: `str2PixelData`?

interface/frontend/src/components/common/OpenDICOM.vue
```diff
@@ -70,27 +70,21 @@
...
       async info () {
         if (typeof this.pool[this.stack.currentImageIdIndex] === 'undefined') {
-          const hola = await this.$axios.get(this.view.prefixUrl + this.view.paths[this.stack.currentImageIdIndex])
+          // workaround for now; sometimes we have full url, sometimes filepath
+          // const hola = await this.$axios.get(this.view.prefixUrl + this.view.paths[this.stack.currentImageIdIndex])
+          var hola
+          if (this.view.paths.length > 0 && this.view.paths[this.stack.currentImageIdIndex].indexOf('=') !== -1) {
+            hola = await this.$axios.get(this.view.paths[this.stack.currentImageIdIndex])
+          } else {
+            hola = await this.$axios.get(this.view.prefixUrl + this.view.paths[this.stack.currentImageIdIndex])
```
reubano: what about the case where `this.view.paths.length == 0`? Should there be an extra`else if` in here?

reubano: also, what about assigning `this.view.paths[this.stack.currentImageIdIndex]` to a var?

interface/backend/images/models.py
```diff
@@ -144,7 +144,8 @@ class ImageSeries(models.Model):
     series_instance_uid = models.CharField(max_length=256)
     uri = models.CharField(max_length=512)
 
-    def get_or_create(uri):
+    @classmethod
+    def get_or_create(cls, uri):
```
lamby: This is definitely a candidate for a model manager!

interface/frontend/src/store.js
```diff
@@ -1,64 +1,79 @@
...
 
-const ROUTE_ORDER = [
-  '/',
-  '/detect-and-select',
-  '/annotate-and-segment',
-  '/report-and-export'
-]
+// we should just use this to get the
+// available endpoints, and then go from there
+const API_ROOT = '/api/'
```
lamby: It would thus be great to not require this to be hardcoded. Possible?

interface/frontend/src/routes.js
```diff
@@ -1,17 +1,28 @@
...
-  { path: '/report-and-export', component: ReportAndExport }
+  {
+    path: '/detect-and-select',
+    component: DetectAndSelect,
+    beforeEnter: (to, from, next) => { if (store.getters.candidatesExist) next() }
+  },
+  {
+    path: '/annotate-and-segment',
+    component: AnnotateAndSegment,
+    beforeEnter: (to, from, next) => { if (store.getters.nodulesExist) next() }
```
lamby: This is neat

interface/backend/cases/models.py
```diff
@@ -140,18 +146,36 @@ class Candidate(models.Model):
...
-    centroid = models.OneToOneField('images.ImageLocation', on_delete=models.CASCADE)
     lung_orientation = models.IntegerField(choices=enums.format_enum(enums.LungOrientation),
                                            default=enums.LungOrientation.NONE)
     appearance_feature = models.IntegerField(choices=enums.format_enum(enums.AppearanceFeature), null=True)
     diameter = models.FloatField(null=True)
     density_feature = models.IntegerField(choices=enums.format_enum(enums.DensityFeature), null=True)
+
+
+@receiver(post_save, sender=Candidate)
+def add_or_remove_nodule_once_candidate_reviewed(sender, instance, *args, **kwargs):
```
lamby: Can we put these into the ``save`` methods? Signals are really magic-at-a-distance with weird error handling semantics :(

isms:
output

PR #265, opened 2017-12-09
Candidate list

isms:
Clean up backend, use store pattern effectively, and add some UI signposts for detect and select screen.

output


interface/frontend/src/components/detect-and-select/CandidateList.vue
```diff
@@ -86,74 +115,64 @@
...
-              } else {
-                console.log('saving review result failed with response: ', response)
-              }
-            })
-            .catch(() => {
-              // TODO: error callback
-            })
+      async addCandidate () {
+        var newCandidate = {
+          centroid: {x: 5, y: 5, z: 5},
```
pjbull: Won't this not exist for some images? Is there a better default?

interface/backend/api/tests.py
```diff
@@ -29,20 +30,23 @@ def test_nodule_list_viewset(self):
...
+        # now create nodules and figure out what we expect to see in the list
         case = CaseFactory()
         candidates = CandidateFactory.create_batch(size=3, case=case)
         nodules = []
         for candidate in candidates:
             nodule, _ = candidate.get_or_create_nodule()
             nodules.append(nodule)
         serialized = [NoduleSerializer(n, context={'request': None}) for n in nodules]
-        expected = [s.data for s in serialized]
+        expecteds = [json.loads(json.dumps(s.data)) for s in serialized]
```
lamby: Hang on, what? :)

interface/backend/cases/tests.py
```diff
@@ -3,12 +3,10 @@
     CaseFactory,
     NoduleFactory
 )
-from backend.cases.models import (
-    Case,
-    Candidate
+from backend.images.factories import (
+    ImageLocationFactory,
+    ImageSeriesFactory
```
lamby: Picky mode - trailing comma? :)

interface/frontend/src/components/detect-and-select/CandidateList.vue
```diff
@@ -11,7 +11,13 @@
                       'candidate-marked': candidate.review_result === REVIEW_RESULT.MARKED}">
                     <span class="candidate-title">
                       Candidate {{ index + 1 }}
-                    </span> (p={{ candidate.probability_concerning.toFixed(3) }})
+                    </span>
+                    <span v-if="candidate.probability_concerning">
```
lamby: When would it be null out of interest?

interface/backend/api/tests.py
isms: This but `.data` of a serialized instance is an `OrderedDict` with non-JSON-primitive objects (e.g. an `enum.Enum` instance rather than `1`), so for comparing values with `assertDictEqual` it was easier to simulate a trip 'round the API (serialized to JSON, then unserialized).

I agree that this is gross but not sure how else to deal with it other than avoiding `assertDictEqual`. Any suggestions would be welcome.

interface/frontend/src/components/detect-and-select/CandidateList.vue
isms: When `added_by_hand` - the `probability_concerning` always comes from the models. You _could_ set it to 1.0 if hand labeled but (a) perfect certainty isn't necessarily a good representation of what the clinician believes, and (b) we already handle human-labeled P(concerning) in the next screen.

interface/backend/api/tests.py
reubano: Perhaps add a comment to that effect to remind future you? :)

tests/test_routes.py
```diff
@@ -1,55 +0,0 @@
...
-                one_failed |= page_errors(base_url + site)
-            services_to_test.remove(service)
-
-        if not services_to_test:
-            sys.exit(1 if one_failed else 0)
-        time.sleep(1)
-
-    closed_ports = [SERVICE_PORTS[service]["PORT"] for service in SERVICE_PORTS.keys()]
-    print("Time limit of {} seconds exceeded. Following ports did not open in time: {}".format(TIMEOUT, closed_ports))
-    sys.exit(1)
```
reubano: Why remove this file?

reubano: https://github.com/concept-to-clinic/concept-to-clinic/issues/160

isms: I removed it on the branch because it was making the builds take forever... Don't necessarily have a problem adding it back but I am not sure what we're testing here. The Docker config? If so is that in scope?

reubano: See above issue.

lamby: Yeah, was quite whle getting that landed... so we should definitely spend a little time before removing it! :)

interface/backend/api/tests.py
lamby: Mmm,. please add as a comment :)

tests/test_routes.py
isms: I commented on #160.

lamby:
How we doing with this PR? :) What are the next steps here?

isms:
I still need to fix the mysteriously failing JS tests.

isms:
:bow:

Issue #240, opened 2017-11-20
DICOM viewer: zoom, pan, and navigate through slices


Serhiy-Shekhovtsov:
According to documentation:

> Allow the user to zoom, pan, and navigate through slices in the image viewer.

During these actions the following model properties should be updated: `zoomRate`,
`offsetX`, `offsetY`. These properties has been introduced in [this commit](https://github.com/concept-to-clinic/concept-to-clinic/pull/226/commits/9e53048dbd65d8d232787e957ba2db2b7f622c48#diff-70d179422138a3042c2135a282065626R42). Hopefully PR will be merged soon. 

These properties should taken into by "Detect and Select" and "Segmentation" components (#226 and #150)

## Current Behavior
User can't navigate through slices, nor zoom+pan.

## Checklist before submitting
- [x] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [x] I provided a **minimal code snippet** or list of steps that reproduces the bug.
- [x] I provided **screenshots** where appropriate
- [x] I filled out all the relevant sections of this template

Serhiy-Shekhovtsov:
Looks like the new cornerstone version have support for zoom and pan. Will check how can we use it with our components for nodule selection and segmentation.

Serhiy-Shekhovtsov:
Activated window width and window center adjustment, zooming and panning of cornerstone tools component. Updated the `NoduleMarker` component to rescale and move the marker accordingly.
![zoom, pan](http://g.recordit.co/VbxLfPX3qB.gif)
Will push as soon as PR #269 is merged.

Serhiy-Shekhovtsov:
The feature has been implemented.

PR #269, opened 2017-12-26
Updated cornerstone-tools, removed jquery-slim dependency

Serhiy-Shekhovtsov:
## Description
Fixed the code to work with new version of `cornerstone-tools`, updated references in `package.json`.

#215 

## CLA


interface/frontend/src/components/common/OpenDICOM.vue
```diff
@@ -128,7 +129,7 @@
         if (_.isEmpty(dicom)) return ''
         else {
           cornerstone.registerImageLoader(this.view.type, () => {
-            return new Promise((resolve) => { resolve(dicom) })
+            return {promise: new Promise((resolve) => { resolve(dicom) })}
```
lamby: How come?

Serhiy-Shekhovtsov: As far as I can see, the most recent version of the tool expects the loader to return an object with two properties: the promise and cancel function. [Here is](https://github.com/cornerstonejs/cornerstoneTools/blob/29182180ed3f850ba41c609b98b96464affca5f0/examples/exampleTextImageLoader.js#L58) an example.

interface/frontend/src/components/common/OpenDICOM.vue
```diff
@@ -128,7 +129,7 @@
         if (_.isEmpty(dicom)) return ''
```
louisgv: Off-topic: I went and search for lodash usage and this is the only thing we're using it for?

https://github.com/concept-to-clinic/concept-to-clinic/search?utf8=%E2%9C%93&q=lodash&type=

Wonder if we should keep lodash :-? ... Most of its feature can be done with ES6

Overall, LGTM

Serhiy-Shekhovtsov: I will clean it up and push together with [these changes](https://github.com/concept-to-clinic/concept-to-clinic/issues/240#issuecomment-354178232), since that file will be changed a lot there.

lamby:
Thanks!

PR #270, opened 2017-12-28
DICOM viewer: zoom and pan support

Serhiy-Shekhovtsov:
Enabled window width and window center adjustment. Enabled zooming and panning. 
Updated the `NoduleMarker` component to rescale and move the marker accordingly.
Fixed bug - moving slider to cached slices didn't show those slices.

#240 

## Screenshot
zoom, pan

## CLA

### DIFF
--- a/interface/frontend/src/components/common/NoduleMarker.vue
+++ b/interface/frontend/src/components/common/NoduleMarker.vue
@@ -1,6 +1,6 @@
 <template>
   <vue-draggable-resizable class="nodule-marker" :class="{ overlapped: markerOverlappedBySlice }"
-                           v-if="scaledMarker" :x="scaledMarker.x" :y="scaledMarker.y"
+                           v-if="dynamicMarker" :x="dynamicMarker.x" :y="dynamicMarker.y"
                            :w="markerImageSize" :h="markerImageSize"
                            :minw="markerImageSize" :minh="markerImageSize"
                            :resizable="false" v-on:dragstop="onDragFinished">
@@ -16,9 +16,7 @@
     components: {VueDraggableResizable},
     name: 'open-dicom',
     props: {
-      zoomRate: null,
-      offsetX: null,
-      offsetY: null,
+      translation: null,
       marker: null,
       sliceIndex: {
         type: Number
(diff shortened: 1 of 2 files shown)


interface/frontend/src/components/common/NoduleMarker.vue
```diff
@@ -38,42 +36,60 @@
...
-        const val = this.scaledMarker.z !== this.sliceIndex
-        return val
-      }
-    },
-    methods: {
       onDragFinished (x, y) {
-        var scaledX = x / this.zoomRate + this.offsetX + this.markerImageSize / 2
-        var scaledY = y / this.zoomRate + this.offsetY + this.markerImageSize / 2
+        var scaledX = x
+        var scaledY = y
```
louisgv: If mutable, `let` should be preferred over `var`. However, I would suggest doing it like so:

```
const scaled = this.translation
    ? {
             x : Math.round((x + this.markerImageSize / 2 - this.translation.offsetX) / this.translation.scale),
             y : Math.round((y + this.markerImageSize / 2 - this.translation.offsetY) / this.translation.scale)
        } 
    : { x, y }
         
    EventBus.$emit('nodule-marker-moved', scaled.x, scaled.y, this.sliceIndex)

```

interface/frontend/src/components/common/OpenDICOM.vue
```diff
@@ -1,20 +1,22 @@
 <template>
-  <div class="DICOM-container" style="width: 100%">
```
lamby: Please don't make unrelated changes in commits - when you use the word "and" or a command (",") in a commit message, it often means that you should split it up into separate commits - makes reviewing and discussion so much easier! (eg. why remove this style? you could explain here but "nobody" will look or can find at this review comment later alas..)

Serhiy-Shekhovtsov: @lamby noted! Will update my PR. Btw, this change is a part of fixing the bug with misinterpretation of the predicted coordinates.

Serhiy-Shekhovtsov: Moved to separate commit.

interface/frontend/src/components/common/NoduleMarker.vue
Serhiy-Shekhovtsov: Done :)

Serhiy-Shekhovtsov:
Looks like the build has been failing since [this PR](https://travis-ci.org/concept-to-clinic/concept-to-clinic/builds/322566530).

lamby:
Thanks!

PR #272, opened 2018-01-03
Fixed coordinates scaling for classification prediction

Serhiy-Shekhovtsov:
We had a problem - the preprocessing is [zooming the image](https://github.com/concept-to-clinic/concept-to-clinic/blob/666fb1bf410cbd80eb7928d0e3c5f2433ad668a5/prediction/src/preprocess/preprocess_ct.py#L144). In other words, the pixel matrix is resized to real proportions. But coordinates we are using for prediction are not scaled accordingly and the actual patch we are making prediction on differs from the expected patch.
For example, the zooming factor for full LIDC-IDRI-0003 image is [2.5, 0.820312, 0.820312]. The prediction result for **coordinates of the real nodule** {'x': 367, 'y': 350, 'z': 72} is **0.0066**. As you can see, it's very low. But if you scale these coordinates by the current zooming factor ({'x': 301, 'y': 286, 'z': 180}) we will get much higher probability of concerning  **0.42**.
More details [here](https://github.com/concept-to-clinic/concept-to-clinic/issues/268#issuecomment-354797905).

## CLA
- [x] I have signed the CLA; if other committers are in the commit history, they have signed the CLA as well

PR #273, opened 2018-01-05
fixed classification prediction for multiple nodules

Serhiy-Shekhovtsov:
Bug fixed. When the length of the nodules list is more then 1 – prediction fails as follows:

```
~/concept-to-clinic/prediction/src/algorithms/classify/src/gtr123_model.py in predict(ct_path, nodule_list, model_path)
    275     results = []
    276 
--> 277     for nodule, (cropped_image, coords) in zip(nodule_list, patches):
    278         cropped_image = Variable(torch.from_numpy(cropped_image[np.newaxis, np.newaxis]).float())
    279         cropped_image.volatile = True

ValueError: too many values to unpack (expected 2)
```

More details [here](https://github.com/concept-to-clinic/concept-to-clinic/pull/272#issuecomment-355413434).

## CLA


reubano:
Thanks for this! Would you also mind adding a test that fails unless this patch is applied? That will make sure we don't create regressions with future changes.

Serhiy-Shekhovtsov:
@reubano, sure. I will make a new PR for this.

Issue #285, opened 2018-01-10
User should be able to run nodule candidates prediction


Serhiy-Shekhovtsov:
Currently we can only create a new case based on local images. There should be a button to start the actual prediction. After clicking it should show that prediction is running. Once prediction is ready - it should show update the status on the page.

## Checklist before submitting
- [ ] I have confirmed this using the officially supported Docker Compose setup using the `local.yml` configuration and ensured that I built the containers again and they reflect the most recent version of the project at the `HEAD` commit on the `master` branch 
- [ ] I have searched through the other currently open issues and am confident this is not a duplicate of an existing bug
- [ ] I provided a **minimal code snippet** or list of steps that reproduces the bug.
- [ ] I provided **screenshots** where appropriate
- [ ] I filled out all the relevant sections of this template

Serhiy-Shekhovtsov:
Guys, if anyone is interested, feel free to take it.

chaitanyarahalkar:
can anyone please explain to me how to use the prediction API? I am in a fix, trying to figure out how can I use it

Serhiy-Shekhovtsov:
You can find the prediction code here: https://github.com/drivendataorg/concept-to-clinic/tree/master/prediction/src/algorithms
And here you can find some usage examples: https://github.com/drivendataorg/concept-to-clinic/tree/master/prediction/src/tests

chaitanyarahalkar:
Hey, thanks for the help! I wanted to know whether it is possible to run the prediction code from the frontend on the sample LIDC images available in the Docker container by default.

chaitanyarahalkar:
Has anyone worked on adding that button as @Serhiy-Shekhovtsov mentioned?

PR #286, opened 2018-01-14
Sending and saving of nodule properties on Annotate screen

Serhiy-Shekhovtsov:
Fixed few bugs of Annotate screen. Fixed sending of the Nodule data to backend. Added fields for storing `probability_concerning` and `note`. Includes fix of #285

## CLA


interface/backend/api/serializers.py
```diff
@@ -95,6 +95,9 @@ class Meta:
     def create(self, validated_data):
         return Nodule.objects.get_or_create(candidate=validated_data['candidate'])
 
+    def update(self, instance, validated_data):
+        return super().update(instance, validated_data)
```
WGierke: Why is this necessary? It doesn't overwrite the update method inherited from the parent class, does it?

Serhiy-Shekhovtsov: Yes, put it while troubleshooting an other issue and forgot to remove. Thanks for noting!
Will fix the PR tomorrow.

Issue #279, opened 2018-01-05
UI feature: add overall progress indicator ("wizard" element)


isms:
## Expected Behavior

We want the users to have a clear path in the workflow:
- Load a case
- Mark all nodules
- Mark any nodules not found if necessary
- Annotate and bound all nodules marked concerning
- Move to report

We should clearly show them at all times a progress indicator showing which phase (step 1/4) and what remains in that phase (2/3 nodules identified).

This should all be fairly easy to calculate from the current Case. Some random images of the kind of element:

![](http://d5vf6134d8ffdnfp1qv4rv3l.wpengine.netdna-cdn.com/wp-content/uploads/06-vertical-progress-checkout.jpg)
![](https://i.gyazo.com/298959b3aacc180e1dce0b27ff65522b.png)
![]()

## Possible implementation

Example: http://vuejs.creative-tim.com/vue-form-wizard/#/

louisgv:
I can start working on this if nobody has started yet.

louisgv:
@Serhiy-Shekhovtsov are you actively working on this issue? (Since you upvoted it). Just wanted to make sure we're not colliding.

Serhiy-Shekhovtsov:
@louisgv I am not working on it, just like the feature :) Especially if the progress will be saved on loaded. This gives me pain every time I work on interface frontend. Have to go to first step and select the case each time the page is reloaded.

louisgv:
I will start working on this. WIP PR coming soon!

PR #288, opened 2018-01-17
[WIP] Wizard flow

louisgv:
## Description

Work in progress wizard flow.

#279 

## Motivation and Context

Adding a wizard flow for each tab to assist user on what they should be doing next.

## CLA

### DIFF
--- a/interface/frontend/package.json
+++ b/interface/frontend/package.json
@@ -43,8 +43,8 @@
     "babel": "inherit"
   },
   "dependencies": {
-    "vue-resource": "^1.3.4",
-    "vue": "^2.5.6"
+    "vue": "^2.5.6",
+    "vue-resource": "^1.3.4"
   },
   "devDependencies": {
     "autoprefixer": "^7.1.2",
@@ -102,6 +102,7 @@
     "shelljs": "^0.7.6",
     "url-loader": "^0.5.8",
     "vue-draggable-resizable": "^1.5.3",
+    "vue-form-wizard": "^0.8.2",
     "vue-loader": "^13.0.4",
     "vue-router": "^2.7.0",
     "vue-style-loader": "^3.0.1",
--- a/interface/frontend/src/components/common/WizardFlow.vue
+++ b/interface/frontend/src/components/common/WizardFlow.vue
@@ -1,37 +1,27 @@
 <template>
-  <!-- @on-complete="onComplete"  -->
-<form-wizard
-  title subtitle
-  shape="tab"
-  color="#453DF4">
-  <tab-content title="Personal details">
-    My first tab content
-  </tab-content>
-  <tab-content title="Additional Info">
-    My second tab content
-  </tab-content>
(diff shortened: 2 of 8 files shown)


lamby:
:+1:

lamby:
Let me know when this is ready for review :

lamby:
Hey, any update on this? :)

louisgv:
@lamby After experimenting with the dependency it seems to me it's not fit for the job (some of the UI element are enforced, like the next/back buttons) so I've been drafting out our own component.

lamby:
Getcha. Shall we close up this PR for the time being? :)

louisgv:
@lamby I'm actually pretty close to have something that I can show. Will ping back status within the weekends.

louisgv:
@lamby Here's some progress:

I created a Wizard Flow component (no styling as of yet) that accept a `steps` computed prop. The steps prop basically tell the component the step and its active state, and will be controlled by the parent component (Which is the view). The parent can either bind the active state by watching the store state, or toggle it manually it it wishes to.

Below demonstrate three working scenario: 

+ If I select an invalid case, none of the wizard are highlighted:

output

+ If I select an imagery, the first wizard will be highlighted:

output

+ If I start a case, the second will be highlighted:
output

+ If the case was created is valid and I select it, all three will be highlighted. (For some reason, I couldn't create a valid case, so no screencaps available)

+ If I select a valid case from the beginning, the first and last will be highlighted:
output


This is a standalone component so implement it for the remaining two should be simple once it is further nailed down.

louisgv:
@lamby @Serhiy-Shekhovtsov please feedback on features you think would be nice to have for this feature. Like onclick binding, metadata, image, step number indicator, subtitle, etc...

isms:
@louisgv I think this is an interesting proof of concept but the layout doesn't seem like something users would be used to. It's not really a wizard like those in the example pictures, and it's not really like the "tour" feature we've seen that guides users through specific steps.

interface/frontend/package.json
```diff
@@ -43,8 +43,8 @@
     "babel": "inherit"
   },
   "dependencies": {
-    "vue-resource": "^1.3.4",
-    "vue": "^2.5.6"
+    "vue": "^2.5.6",
+    "vue-resource": "^1.3.4"
```
isms: This project is using Axios instead of Vue Resource.

louisgv: vue-resoure as a dep was there from the beginning. I can remove it once we confirm there are no trace of its usage.

louisgv:
@isms I see! If that's the case, I imagine this wizard would be more of a horizontal element that vertical?
If so, we might be able to utilize the vue wizard dependency (tho, some of its UI might still not make sense for our app), or I can create a common Flow component extending what I have that will observe the global cross-page steps.

isms:
@louisgv I'm going to go ahead and close this, but we'll still recognize the effort. Thanks for working on it.

PR #289, opened 2018-01-18
/<algorithm>/predict/: Only expect GET or POST requests

WGierke:
While skimming through [code that still needs to be tested in the prediction service](https://wgierke.github.io/stuff/prediction/index.html), I noticed this line that prevented the file from being completely covered by tests. The reason was that the condition was never false since the `methods=['GET', 'POST']` parameter prevents any other request than a `POST` request from reaching this LOC.

## CLA


lamby:
Thanks, and good spot!

Issue #271, opened 2018-01-03
Evaluation Pipeline for Models


WGierke:
As discussed in #131 it would be helpful to have a consistent pipeline to evaluate prediction models. This way we get to know how well the currently implemented models are, which ones need to be improved and how well a new model performs. The pipeline should calculate the appropriate metrics that have been specified in #221 while some of the [are already available here](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/prediction/src/algorithms/segment/src/evaluate.py).

## Acceptance Criteria
- [ ] all evaluation methods take a model and a parameter whether to use 5-fold validation or a test set (default)
- [ ] for classification calculate accuracy, Log Loss
- [ ] for identification calculate accuracy, Log Loss
- [ ] for segmentation calculate Dice coefficient, Hausdorff distance, Jaccard index, sensitivity, specificity
- [ ] for all of them calculate data IO, disk space usage, memory usage, prediction time (if not (easily) possible, specify why and how to manually measure it)
- [ ] for all of them calculate the training time

Please refer to [the PR template](https://github.com/concept-to-clinic/concept-to-clinic/blob/master/.github/PULL_REQUEST_TEMPLATE.md) for further explanations of the metrics.

vessemer:
For the both identification and classification (false positive reduction) tasks was proposed a handy evaluation [framework](https://luna16.grand-challenge.org/evaluation/) by the LUNA16 authors. 
They employed Free-Response Receiver Operating Characteristic ([FROC](https://doi.org/10.1093/jicru/ndx009)) and competition performance metric ([CPM](https://doi.org/10.1109/TMI.2010.2072789)). It computes an average of the seven sensitivities measured at several false positives per scan (FPPS) thresholds, more concretely, at each FPPS ∈ {0.125, 0.25, 0.5, 1, 2, 4, 8} true positive rate was computed. Mean of which forms the CPM. From my point of view, it worth to pay attention to the CPM neither the logloss.
I can work on that to adjust their pipeline, if no one mind.

reubano:
@vessemer great observation! I see that they provide evaluation code as well. So yes, adjusting it to fit our use case will be extremely useful!

PR #290, opened 2018-01-19
#271 Classification Evaluation

WGierke:
I'm currently adding the possibility to evaluate a classification model based on the LIDC dataset as described in #271. Furthermore, I'm currently benchmarking the model that's implemented at the moment.
Even if I still haven't finished, I wanted to show you what I'm currently working at.

## CLA

### DIFF
--- a/prediction/requirements/base.txt
+++ b/prediction/requirements/base.txt
@@ -10,3 +10,4 @@ opencv-python==3.3.0.10
 pandas==0.20.3
 scikit-image==0.13.0
 SimpleITK==1.0.1
+tqdm==4.19.5
--- a/prediction/src/algorithms/classify/src/gtr123_model.py
+++ b/prediction/src/algorithms/classify/src/gtr123_model.py
@@ -208,7 +208,6 @@ class CaseNet(nn.Module):
         """
         xsize = xlist.size()
         corrdsize = coordlist.size()
-        print(xsize)
         # xlist = xlist.view(-1,xsize[2],xsize[3],xsize[4],xsize[5])
         # coordlist = coordlist.view(-1,corrdsize[2],corrdsize[3],corrdsize[4],corrdsize[5])
 
--- a/prediction/src/algorithms/evaluation/evaluation.py
+++ b/prediction/src/algorithms/evaluation/evaluation.py
@@ -2,17 +2,18 @@
 This file contains logic to evaluate the prediction models.
 """
 import os
-import sys
-
-sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', '..')))
 
 import numpy as np
 import pylidc as pl
-from config import Config
-from src.algorithms.classify import trained_model
 from tqdm import tqdm
-from src.preprocess.lung_segmentation import get_z_range
+from src.algorithms.classify import trained_model
 from src.algorithms.evaluation.metrics import get_accuracy, logloss
+from src.preprocess.lung_segmentation import get_z_range
+
+try:
+    from ....config import Config
+except ValueError:
+    from config import Config
 
 CONFIDENCE_THRESHOLD = 0.5
 
--- a/prediction/src/algorithms/evaluation/evaluation.py
+++ b/prediction/src/algorithms/evaluation/evaluation.py
@@ -1,0 +1,58 @@
+"""
+This file contains logic to evaluate the prediction models.
+"""
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', '..')))
+
+import numpy as np
+import pylidc as pl
+from config import Config
+from src.algorithms.classify import trained_model
+from tqdm import tqdm
+from src.preprocess.lung_segmentation import get_z_range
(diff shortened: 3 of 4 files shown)


prediction/src/algorithms/evaluation.py
```diff
@@ -0,0 +1,58 @@
...
+    log_losses = []
+
+    for scan in tqdm(scans[:5]):
+        dicom_path = os.path.join(Config.SMALL_DICOM_PATHS, scan.patient_id, scan.study_instance_uid,
+                                  scan.series_instance_uid)
+        min_z, max_z = get_z_range(dicom_path)
+        nodule_list = []
+        for annotation in scan.annotations:
+            centroid_x, centroid_y, centroid_z = annotation.centroid()
+            z_index = int((centroid_z - min_z) / scan.slice_thickness)
```
WGierke: This step should maybe be transferred to an other location where it can also be tested separately. Computing the slice index from the relative image positions still feels a bit hacky.

WGierke:
The current implementation takes 8.5h to run, gives an average accuracy/precision of 10% and an average loss of 3.55 :/

prediction/src/algorithms/evaluation.py
WGierke: @Serhiy-Shekhovtsov @vessemer Do you have experience with getting the slice index from the ImagePositient information? The DICOM images have ImagePosition information (z_axis) from e.g. -435 to -63. The nodule has one of -252 so we're required to calculate the slice index (here: 61) out of that. Do you know whether that's the right way or whether we are already doing this somewhere?

Serhiy-Shekhovtsov: Is [this](https://github.com/concept-to-clinic/concept-to-clinic/blob/666fb1bf410cbd80eb7928d0e3c5f2433ad668a5/prediction/src/preprocess/load_ct.py#L185) what you are looking for?

Serhiy-Shekhovtsov: The offset then used [here](https://github.com/concept-to-clinic/concept-to-clinic/blob/413bc74311579bc3eaa9ba9a147eb8bdc937b093/prediction/src/preprocess/preprocess_ct.py#L167) to convert the coordinates

```
    coord = np.rint((coord - np.array(origin)) * np.array(spacing))
```

prediction/src/algorithms/evaluation.py
```diff
@@ -0,0 +1,58 @@
...
+def logloss(true_label, predicted, eps=1e-15):
+    """Calculate the logarithmic loss (http://wiki.fast.ai/index.php/Log_Loss)"""
+    p = np.clip(predicted, eps, 1 - eps)
+    if true_label == 1:
+        return -np.log(p)
+    else:
+        return -np.log(1 - p)
+
+
+def evaluate_classification(model_path=None):
```
vessemer: Will it be better to make evaluation functionality console executable with appropriate metrics and directories arguments? Mean, that this is a bit out of user usage. 
I'd also suggest to put all metrics in a different file.

prediction/src/algorithms/evaluation.py
```diff
@@ -0,0 +1,58 @@
...
+import pylidc as pl
+from config import Config
+from classify import trained_model
+from tqdm import tqdm
+from src.preprocess.lung_segmentation import get_z_range
+
+CONFIDENCE_THRESHOLD = 0.5
+
+
+def get_accuracy(TP, TN, FP, FN):
```
reubano: would you be able to use more descriptive variable names?

WGierke: Sure. Do you maybe see any major flaws in the implemented logic of the whole file? I'm just asking because an accuracy of 10% seems pretty low to me.

reubano:
It looks good, and no I don't see any major flaws. If someone with more experience in applying these metrics to CT scans has suggestions he/she can always improve this with future PRs. It also seems like the `sys.path.insert...` isn't making Travis happy. I'll see what can be done about that.

reubano:
This is what I got...

`prediction/src/algorithms/evaluation/evaluation.py`

```python
import os

import numpy as np
import pylidc as pl

try:
    from ....config import Config
except ValueError:
    from config import Config

...
```

`src/tests/test_evaluate_classification.py`

```python
from ..algorithms.evaluation.evaluation import evaluate_classification


def test_evaluate_classification(model_path=None):
    assert evaluate_classification(model_path)
```

run with `docker-compose -f local.yml run prediction pytest src/tests/test_evaluate_classification.py`

I believe you'll also have to add `tdmq` to the requirements and fix whatever remaining style errors you may see via `flake8 prediction`

WGierke:
@reubano Done. Thanks!

lamby:
Awesome :)

prediction/src/algorithms/evaluation.py
vessemer: @WGierke, @Serhiy-Shekhovtsov, I've missed that, it should be `np.rint((coord - np.array(origin)) / np.array(spacing))`, also I didn't get for what reason, @Serhiy-Shekhovtsov, you remove [this](https://github.com/concept-to-clinic/concept-to-clinic/blame/a2b55a863877ac790292969ce072903d552f3abc/prediction/src/preprocess/preprocess_ct.py#L145) line, I should missing something :/

PR #292, opened 2018-01-22
#271 LUNA16 evaluation

vessemer:
This allows to compute [FROC](https://doi.org/10.1093/jicru/ndx009) and [CPM](https://doi.org/10.1109/TMI.2010.2072789) metrics with bootstrapping for the classification and detection tasks.

## Description
Free-Response Receiver Operating Characteristic ([FROC](https://doi.org/10.1093/jicru/ndx009)) and competition performance metric ([CPM](https://doi.org/10.1109/TMI.2010.2072789)). It computes an average of the seven sensitivities measured at several false positives per scan (FPPS) thresholds, more concretely, at each FPPS ∈ {0.125, 0.25, 0.5, 1, 2, 4, 8} true positive rate was computed. Mean of which forms the CPM as discussed in #271. 

## Reference to official issue
This reference to the #271.

## How Has This Been Tested?  
froc_3dlrcnn

[CPM](https://doi.org/10.1109/TMI.2010.2072789) over 10-Fold cross validation:

| 0.125 | 0.25 | 0.5 | 1 | 2 | 4 | 8 | Score ([CPM](https://doi.org/10.1109/TMI.2010.2072789)) |
|:-----:| ----- | ----- | ----- | ----- | ----- | ----- |:-----:|
| 0.595 | 0.670 | 0.731 |  0.793 |  0.835 | 0.868 | 0.887 | 0.76 |

## CLA
- [x] I have signed the CLA; if other committers are in the commit history, they have signed the CLA as well

Issue #131, opened 2017-09-20
Continuous improvement of nodule classification models (see #2)


isms:
## Overview

We want to continuously improve the accuracy and reliability of models developed for nodule classification. This is a continuation of #2 and will remain open indefinitely.

**Note:** Substantive contributions are currently eligible for increased point awards.

Design doc reference:
[Jobs to be done > Detect and select > Prediction service](https://concept-to-clinic.readthedocs.io/en/latest/design-doc.html#prediction-service)

## Acceptance criteria

 - [ ] trained model for classification
 - [ ] documentation for the trained model (e.g., cross validation performance, data used) and how to re-train it

*NOTE: All PRs must follow the standard PR checklist.*

WGierke:
Maybe we should proceed in small steps by e.g. first adding evaluation methods that cover most of the metrics introduced in #221 so we can have a more standardized way to determine those metrics? This way we would also be able to quickly determine the quality of the current implementations of the identification, classification and segmentation algorithms. This would in turn make it easier to focus first on the algorithm that is performing the worst so far. Any thoughts @isms @reubano @lamby ?

pjbull:
For folks who are interested, see the latest announcement here:
https://concepttoclinic.drivendata.org/newsfeed

There are a limited number of AWS credits available for folks to continue to make progress on these algorithms.

WGierke:
Thanks @pjbull !
I contacted the DrivenData team and asked for credits. If they can provide me with some, I'd like to work on this issue. Once I get their answer, I'll update this comment with the current state for transparency reasons.

Serhiy-Shekhovtsov:
@WGierke, I have create a virtual machine, the support should increase the instances limit anytime soon, and then the machine will be ready for use. Also, I have plenty of time for upcoming week. So, if you have some ideas, we can work together on the issue and share the points and the fun :)
If this sounds good to you, please, [contact me](https://gitter.im/Serhiy-Shekhovtsov) on Gitter.

Serhiy-Shekhovtsov:
@reubano, @pjbull what @WGierke [said here](https://github.com/concept-to-clinic/concept-to-clinic/issues/131#issuecomment-351563298) makes perfect sense to me. If the score for those metrics is a required part of models improvement it would be nice to have a set of standard tests for them.
I am happy to take part in it's development. What do you think?

reubano:
[...]

PR #298, opened 2018-01-24
Classification model pipeline

vessemer:
Pipeline to train and predict by the model was provided along with an interface for `classification_model`. 

#131 

## Metrics:
Model' CPM score was described in PR #292:

[CPM](https://doi.org/10.1109/TMI.2010.2072789) over 10-Fold cross validation:

| 0.125 | 0.25 | 0.5 | 1 | 2 | 4 | 8 | Score ([CPM](https://doi.org/10.1109/TMI.2010.2072789)) |
|:-----:| ----- | ----- | ----- | ----- | ----- | ----- |:-----:|
| 0.595 | 0.670 | 0.731 |  0.793 |  0.835 | 0.868 | 0.887 | 0.76 |   

froc_3dlrcnn  

| Item | Config |
| :---: | :----: |
| GPU | NVIDIA TITAN X |
| GPU memory usage | 12GiB for batch size 32 | 

## CLA


vessemer:
Currently, I'm working on the training part adjustment of `grt123` algorithm.

prediction/src/algorithms/classify/src/classification_model.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+
+    Args:
+        init_model (bool | str): whether to initialise the model.
+            If str, then the model will be loaded from the init_model
+            path.
+        pull_size (int): maximum amount of batches allowed to be stored in RAM.
+    """
+
+    @abc.abstractmethod
```
reubano: what does `abstractmethod` do in this case? Python docs say, 

> Using this decorator requires that the class’s metaclass is ABCMeta or is derived from it. 

Also, why not use a new style class, i.e., `class ClassificationModel(object):`

reubano:
Restarted the travis build since the error occurred before even reaching the tests.

vessemer:
Okh, [this](https://github.com/concept-to-clinic/concept-to-clinic/blame/413bc74311579bc3eaa9ba9a147eb8bdc937b093/prediction/src/preprocess/preprocess_ct.py#L167) line along with removal [this](https://github.com/concept-to-clinic/concept-to-clinic/blame/fc9137c2c110f035cdd068bf4b2a5ec479357fc2/prediction/src/preprocess/preprocess_ct.py#L111) one makes me suffer for a while and force to code and tests updates. Unfortunately, at the moment of PR #272 I did not have the opportunity to provide a review, now to be clear:
First: `ndarray` coordinates of the real world point should be computed in this way: `(point - origin) / spacing`and not [`(point - origin) * spacing`](https://github.com/concept-to-clinic/concept-to-clinic/blame/fc9137c2c110f035cdd068bf4b2a5ec479357fc2/prediction/src/preprocess/preprocess_ct.py#L111), where `spacing` is the shape of one voxel in real-world units. 
Second: `meta.spacing` stores information of aforementioned voxel's shape, and if we apply affine transformation such as zoom, then `meta.spacing` should changes too, which is not the previous behaviour. The way that `preprocess_ct.PreprocessCT` handle `spacing` parameter is to zoom an ndarray exactly to this new `spacing` by setting `zoom_fctr` to be `old_spacing / new_spacing` this means that we need no more to store the `old_spacing` in `meta` and at the same time it's enough to store only current spacing for real-world - ndarray coordinates translations. That's why [this](https://github.com/concept-to-clinic/concept-to-clinic/blame/fc9137c2c110f035cdd068bf4b2a5ec479357fc2/prediction/src/preprocess/preprocess_ct.py#L111) line is important.
This mistakes were propagated to the tests in a way of coordinates picking.

Quick check:
tmp

prediction/src/algorithms/classify/src/classification_model.py
vessemer: Indeed, thanks :)

lamby:
@vessemer Can you rework your previous comment into the code itself? Remarks here are really really useful (!) but it won't be clear to anyone following the codebase later, you see...

vessemer:
@lamby, good point, done :)

lamby:
Thanks!

prediction/src/tests/test_classification.py
```diff
@@ -1,24 +1,24 @@
...
-
+# def test_classify_predict_load(metaimage_path, model_path):
+#     assert not trained_model.predict(metaimage_path, [], model_path)
+#
+#
+# def test_classify_dicom(dicom_paths, nodule_locations, model_path):
+#     predicted = trained_model.predict(dicom_paths[0], nodule_locations, model_path)
+#     assert predicted
+#     assert 0 <= predicted[0]['p_concerning'] <= 1
+#
```
reubano: Why comment out these tests? Anything else needed to keep them passing?

vessemer: Not at all, just forget to uncomment them, done [here](https://github.com/concept-to-clinic/concept-to-clinic/pull/304)

PR #305, opened 2018-01-31
[WIP] Refactor imagery loading EventEmitter into store value

louisgv:
## Description

I refactored the event emitter inside the TreeView to send it to the global store. This way if any other component need access to the imagery, it can simply grab it from the store.

## Reference to official issue

#302 

## How Has This Been Tested?

Manual testing

## CLA
- [ ] I have signed the CLA; if other committers are in the commit history, they have signed the CLA as well

PR #309, opened 2018-02-27
Improve various code styles

lamby:
This PR includes a large number of code refactoring within the backend (Django) part of the application.

### DIFF
--- a/interface/backend/api/urls.py
+++ b/interface/backend/api/urls.py
@@ -21,8 +21,10 @@ from rest_framework.urlpatterns import format_suffix_patterns
 
 
 class RelativeUrlRootView(routers.APIRootView):
-    """ Provides relative URLs for the available endpoints.
     """
+    Provides relative URLs for the available endpoints.
+    """
+
     def get(self, request, *args, **kwargs):
         # Return a plain {"name": "hyperlink"} response.
         ret = OrderedDict()
--- a/interface/backend/api/urls.py
+++ b/interface/backend/api/urls.py
@@ -1,4 +1,4 @@
-from collections import OrderedDict
+import collections
 
 from backend.api.views import (
     CaseViewSet,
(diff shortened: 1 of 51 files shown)


lamby:
@isms @reubano  Any ideas on this testsuite failure? Can't seem to work out what is actually really broken, let alone how I could have done it...

reubano:
Possibly related to https://github.com/concept-to-clinic/concept-to-clinic/pull/305#issuecomment-363400089

isms:
@reubano Can you dig in and find out what the source is?

@lamby I say we let that tests fail for a while while Reuben looks into it so you're not blocked :)

lamby:
As in, I can't *quite* understand the error message!

reubano:
> @reubano Can you dig in and find out what the source is?

@isms @lamby can either of you verify that bug?

lamby:
@isms  I'm not blocked, more that I wouldn't want to be *introducing* any new issue. Indeed, this PR can be merged (renaming to match)

interface/backend/api/urls.py
```diff
@@ -1,4 +1,4 @@
-from collections import OrderedDict
+import collections
```
WGierke: Why do you prefer to import the whole module here while you changed in 5302cfe401a1f77fa878fd8ad1bc26dd2d77e224 to directly import methods from a module?

lamby: So in general I prefer to import the modules that Python ships with ("batteries included") as top-level modules. However, I am not 100% dogmatic on this (!) and make some exceptions if that's going to make the code unspeakably ugly. For example,, imagine trying to parse:
```
os.path.join(os.path.join(os.path.join(os.path.ditpath(os.path.join(os.path.join(os.path.join(....)))
```
Make sense? :)

lamby:
Any luck all? Would love to get this merged! :)

interface/backend/api/urls.py
WGierke: Yes, thank you :)

